PR #1, opened 2024-09-15
Code quality things

dvx76:
* Enable (almost) all ruff rules, but disable the most obnoxious or incompatible ones.

I like to start with all rules enabled and then disable anything we agree are overly pedantic or unnecessary.

Fix any remaining findings in the code. Mostly line length and type annotations.

100% open for discussion.

* Add pre-commit config for ruff and fix ruff findings

### DIFF
--- a/myskoda/authorization.py
+++ b/myskoda/authorization.py
@@ -1,20 +1,20 @@
 """Handles authorization to the MySkoda API."""
 
 import base64
-from datetime import datetime, timedelta
 import hashlib
 import json
 import logging
 import random
 import re
 import string
-from typing import cast
 import uuid
+from datetime import UTC, datetime, timedelta
+from typing import cast
 
-from aiohttp import ClientSession, FormData
-from bs4 import BeautifulSoup
 import jwt
 import yaml
+from aiohttp import ClientSession, FormData
+from bs4 import BeautifulSoup
 
 from .const import BASE_URL_IDENT, BASE_URL_SKODA, CLIENT_ID
 
@@ -30,7 +30,7 @@ class IDKCredentials:
     email: str
     password: str
 
-    def __init__(self, data: dict, email: str, password: str):
+    def __init__(self, data: dict, email: str, password: str) -> None:
         """Create a new set of credentials.
 
         The dict is supposed to contain the hmac, csrf and relay state from the website.
@@ -39,9 +39,8 @@ class IDKCredentials:
         self.password = password
         self.update(data)
 
-    def update(self, data: dict):
+    def update(self, data: dict) -> None:
         """Update the values from the website in case the changed."""
-
         self.csrf = cast(str, data.get("csrf_token"))
         self.hmac = cast(str, data.get("templateModel", {}).get("hmac"))
         self.relay_state = data.get("templateModel", {}).get("relayState")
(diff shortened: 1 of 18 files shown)


myskoda/authorization.py
```diff
@@ -172,14 +167,12 @@ async def _initial_oidc_authorize(
         "redirect_uri": "myskoda://redirect/login/",
         "response_type": "code id_token",
         # OpenID scopes. Can be found here: https://identity.vwgroup.io/.well-known/openid-configuration
-        "scope": "address badge birthdate cars driversLicense dealers email mileage mbb nationalIdentifier openid phone profession profile vin",
+        "scope": "address badge birthdate cars driversLicense dealers email mileage mbb nationalIdentifier openid phone profession profile vin",  # noqa: E501
```
WebSpider, line 170: Let's not add an exemption for a line too long if we can easily make it a concatenated string:
"string" \
"string2"

myskoda/authorization.py
```diff
@@ -172,14 +167,12 @@ async def _initial_oidc_authorize(
...
         "code_challenge": challenge,
         "code_challenge_method": "s256",
         "prompt": "login",
     }
-    async with session.get(
-        f"{BASE_URL_IDENT}/oidc/v1/authorize", params=params
-    ) as response:
+    async with session.get(f"{BASE_URL_IDENT}/oidc/v1/authorize", params=params) as response:
```
WebSpider, line 175: I actually don't think this results in better readable code :/

myskoda/rest_api.py
```diff
@@ -182,24 +179,20 @@ async def get_vehicle(self, vin) -> Vehicle:
...
 
-    async def _headers(self):
-        return {
-            "authorization": f"Bearer {await self.idk_session.get_access_token(self.session)}"
-        }
+    async def _headers(self) -> dict[str, str]:
+        return {"authorization": f"Bearer {await self.idk_session.get_access_token(self.session)}"}
 
-    async def stop_air_conditioning(self, vin):
+    async def stop_air_conditioning(self, vin: str) -> None:
```
WebSpider, line 187: General remark: Why do the calls that make a change, return no status? It would seem relevant to know if the request went through or not, unless this information is lost due to the async nature.

pyproject.toml
```diff
@@ -13,8 +13,8 @@ pyjwt = "^2"
 pyyaml = "^6"
 pydantic = "^1"
 asyncio = "^3"
-asyncclick = {version = "^8.1.7.2", optional = true}
-termcolor = {version = "^2.4.0", optional = true}
+asyncclick = { version = "^8.1.7.2", optional = true }
```
WebSpider, line 16: Is it possible to not require an exact version, but rather a minimum/maximum version?

WebSpider:
Oh yeah, I fully agree with this approach! Did a few inline remarks, nothing huge.

Prior99:
Merged this! The only thing that I added was pydantic to `authorization.py`

Prior99:
Thank you for the good work :)

myskoda/authorization.py
dvx76: I think it's just auto-formatting after bumping line-length to 100

myskoda/authorization.py
dvx76: Sounds good, will do in another set of changes.

myskoda/rest_api.py
dvx76: I would only return something if there's something interesting in the response (e.g. number of requests remaining of something like that). If the response of the API call is just a 200-OK then returning None is fine I think.

However maybe in all these methods which just wrap an API call we do want to check the return code and raise an exception if it's not OK...

pyproject.toml
dvx76: Suggest to discuss this on @Prior99 on Discord. Personally I like fairly strict version pinning but with poetry it matters less since everything is already pinned in the lock file.

PR #6, opened 2024-09-18
Drop and log unknown capabilities, log responses, deserialize with Pydantic

dvx76:
Logging API responses will help debugging issues reported by users, like https://github.com/skodaconnect/homeassistant-myskoda/issues/18

I started working on replacing pydantic with mashumaro but I'm not convinced yet. Pydantic does a lot of nice things and does them elegantly, like this additional validator which will drop unknown capabilities. It also enforces types and even coerces 'wrong' types into the desired type. So these validators to convert str to int aren't actually required. Use pydantic (`.parse_raw()`) to de-serialize the json responses.

### DIFF
--- a/myskoda/models/info.py
+++ b/myskoda/models/info.py
@@ -1,9 +1,12 @@
 """Models for responses of api/v2/garage/vehicles/{vin}."""
 
+import logging
 from datetime import date
 from enum import StrEnum
 
-from pydantic import BaseModel, Field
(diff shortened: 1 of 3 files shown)


dvx76:
Just noticed this includes/obsoletes #3

WebSpider (approved):
lgtm

WebSpider:
Yeah basically the same thing as #3

Prior99 (approved):

PR #12, opened 2024-09-22
Add workflow to publish to PyPI

dvx76:
Auto build and publish the package to PyPI when a release is created in Github. We had this in the old project but based on flit instead of poetry. 

Requires a one-time setup to authorize the project in PyPI: https://github.com/marketplace/actions/pypi-publish#trusted-publishing

@Prior99 can you please add @WebSpider and myself as maintainers in PyPI?

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,43 @@
+---
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish-to-pypi:
+    name: Publish Python distribution to PyPI
+    needs:
+      - build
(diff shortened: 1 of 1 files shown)


WebSpider:
Maybe its a good idea to put this in the skodaconnect pypi org?

dvx76:
> Maybe its a good idea to put this in the skodaconnect pypi org?

There is no org in PyPI 😅  . We just have the https://pypi.org/project/skodaconnect/ project.

WebSpider:
Wait, what, why is there no org 😄

joostlek:
Pypi doesn't have orgs

Prior99 (approved):

PR #15, opened 2024-09-22
Add SUV Coupe BodyType

dvx76:
Fixes skodaconnect/homeassistant-myskoda#18


dvx76:
> Lgtm, but i think we should test seperate vehicles, not all of them in one go.
> 
> Seperate vehicles at seperate configurations/software likely have different capabilities.

I agree but this is the trace I got from https://github.com/skodaconnect/homeassistant-myskoda/issues/18 so figured I might as well include it. As we get more traces from various vehicles and endpoints we can refine the tests.

Issue #13, opened 2024-09-22
Failed to refresh access token


Prior99:
Even with the new implementation it will fail:

```
2024-09-22 06:28:09.792 INFO (MainThread) [myskoda.authorization] Refreshing IDK access token
2024-09-22 06:28:09.894 WARNING (MainThread) [myskoda.authorization] Retrying failed request to refresh token.
2024-09-22 06:28:09.917 WARNING (MainThread) [myskoda.authorization] Retrying failed request to refresh token.
2024-09-22 06:28:09.941 WARNING (MainThread) [myskoda.authorization] Retrying failed request to refresh token.
2024-09-22 06:28:09.966 WARNING (MainThread) [myskoda.authorization] Retrying failed request to refresh token.
2024-09-22 06:28:09.987 WARNING (MainThread) [myskoda.authorization] Retrying failed request to refresh token.
2024-09-22 06:28:10.009 ERROR (MainThread) [custom_components.myskoda.coordinator] Unexpected error fetching myskoda data
Traceback (most recent call last):
  File "/home/prior/repositories/myskoda/myskoda/authorization.py", line 83, in perform_refresh
    raise InvalidStatusError(response.status)  # noqa: TRY301
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
myskoda.authorization.InvalidStatusError: Received invalid HTTP status code 400.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/prior/repositories/myskoda/myskoda/authorization.py", line 83, in perform_refresh
    raise InvalidStatusError(response.status)  # noqa: TRY301
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
myskoda.authorization.InvalidStatusError: Received invalid HTTP status code 400.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/prior/repositories/myskoda/myskoda/authorization.py", line 83, in perform_refresh
    raise InvalidStatusError(response.status)  # noqa: TRY301
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
myskoda.authorization.InvalidStatusError: Received invalid HTTP status code 400.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/prior/repositories/myskoda/myskoda/authorization.py", line 83, in perform_refresh
[...]

PR #21, opened 2024-09-22
Fix IDKSession field aliases

dvx76:
Fixes skodaconnect/myskoda#13

🙃

### DIFF
--- a/myskoda/authorization.py
+++ b/myskoda/authorization.py
@@ -65,8 +65,8 @@ class IDKSession(BaseModel):
     """
 
     access_token: str = Field(None, alias="accessToken")
-    refresh_token: str = Field(None, refreshToken="accessToken")
-    id_token: str = Field(None, idToken="accessToken")
+    refresh_token: str = Field(None, alias="refreshToken")
+    id_token: str = Field(None, alias="idToken")
 
     async def perform_refresh(self, session: ClientSession, attempt: int = 0) -> None:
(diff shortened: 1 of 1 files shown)


WebSpider (dismissed):
Great find!

myskoda/authorization.py
```diff
@@ -65,8 +65,8 @@ class IDKSession(BaseModel):
     """
 
     access_token: str = Field(None, alias="accessToken")
-    refresh_token: str = Field(None, refreshToken="accessToken")
-    id_token: str = Field(None, idToken="accessToken")
+    refresh_token: str = Field(None, refreshToken="refreshToken")
+    id_token: str = Field(None, idToken="idToken")
```
Prior99, line 69: Shouldn't it be `alias="refreshToken"`?

dvx76: How do you mean? The initial login response which we parse into this model looks like `{"accessToken": "...", "refreshToken": "...", "idToken": "..."}` so I assume we'd want to store `idToken` in `self.id_token` 😄 

That being said, it doesn't look we're `id_token` is being used for anything (yet?).

dvx76: Derp

dvx76: Fixing it

WebSpider (approved):
looks even better to me :)

Prior99:
I included this in  #10

Issue #20, opened 2024-09-22
Add Suberb test fixture


dvx76:
From skodaconnect/myskoda#19

fursov:
I added PR #23 for this ticket

PR #19, opened 2024-09-22
Update enumerations for Skoda Superb iV 2020

fursov:
Added capabilities and body type ("Liftback")

Based on following response:

```
{
    "vin": "XXXXXXXXXX",
    "name": "Superb",
    "workshopModeEnabled": false,
    "state": "ACTIVATED",
    "devicePlatform": "MBB_ODP",
    "specification": {
        "title": "Škoda Superb",
        "manufacturingDate": "2020-06-08",
        "model": "Superb",
        "modelYear": "2020",
        "body": "Liftback",
        "trimLevel": "L&K iV",
        "systemCode": "UNKNOWN",
        "systemModelId": "3V35XC",
        "battery": {
            "capacityInKWh": 13
        },
        "engine": {
            "type": "TSI iV",
            "powerInKW": 160,
            "capacityInLiters": 1.4
        },
        "gearbox": {
            "type": "A6F"
        }
    },
    "servicePartner": {
        "servicePartnerId": "XXXXXX"
    },
    "renders": [
        {
            "url": "https://iprenders.blob.core.windows.net/base3v3s20201119/0F0FEKAcM-837-icSUg3f_sXOJRZCyGa-CK2vH0E59-4A6othXxnP8dOzuUaqJV-H9AsCqv16YbLX2EKOzn5fjwk-IaBMLRAHXUTFO-19201080dayvext_front1080.png",
            "viewPoint": "exterior_front",
            "type": "REAL",
            "order": 0
        },
        {
            "url": "https://iprenders.blob.core.windows.net/base3v3s20201119/0F0FEKAcM-837-icSUg3f_sXOJRZCyGa-CK2vH0E59-4A6othXxnP8dOzuUaqJV-H9AsCqv16YbLX2EKOzn5fjwk-IaBMLRAHXUTFO-19201080dayvext_side1080.png",
            "viewPoint": "exterior_side",
            "type": "REAL",
            "order": 1
        },
        {
            "url": "https://ip-modcwp.azureedge.net/modcwp3v3s20201119/0F0FEKAcM-837-icSUg3f_sXOJRZCyGa-CK2vH0E59-4A6othXxnP8dOzuUaqJV-H9AsCqv16YbLX2EKOzn5fjwk-IaBMLRAHXUTFO-1920658dayvext_side1080.png?v=638547362740000000",
            "viewPoint": "garage_l",
            "type": "REAL",
            "order": 2
        },
        {
[...]

PR #23, opened 2024-09-23
Add test for Superb info

fursov:
Related to change #19, should have been done it there :)


fursov:
Seems there is something else to fix. I will have a look on this later.

WebSpider:
Fixes #20

PR #27, opened 2024-09-23
Wrap validations in try/except so we can emit some logging

### LINKED ISSUE #22: ValidationErrors should log the source json which failed to be parsed/validated
So we don't rely as much on full tracing/debug being enabled.

WebSpider:
Fixes #22


Prior99 (approved):

WebSpider:
Conflicts with #31, let's see how we can catch validationerrors in mashu

dvx76:
> Conflicts with #31, let's see how we can catch validationerrors in mashu

We may not need too sincee mashu spits out failures like in skodaconnect/homeassistant-myskoda#34

Prior99:
In the meantime I had added this:

```python
    async def _make_get_request[T](self, url: str, deserialize: Callable[[str], T]) -> T:
        async with self.session.get(
            f"{BASE_URL_SKODA}/api{url}",
            headers=await self._headers(),
        ) as response:
            response_text = await response.text()
            try:
                data = deserialize(response_text)
            except Exception:
                _LOGGER.exception(
                    "Failed to load data from url %s. Return value was '%s'", url, response_text
                )
                raise
            else:
                return data
```

Which already wraps all get requests.

Is this PR still needed?

If yes, can we find a more generic way?

dvx76:
I think we can close this. If we still need more/better logs it will bubble up again while looking at issues.

PR #30, opened 2024-09-23
Add doors opened state for Superb

fursov:
Added test for vehicle status call based on the Superb iV responses.

Superb iV also reports OPENED state when any door is opened.


Prior99 (approved):

Prior99:
Awesome. Thank you.

PR #31, opened 2024-09-23
Replace pydantic with mushimushi

Prior99:


myskoda/auth/authorization.py
```diff
@@ -7,20 +7,23 @@
...
 from asyncio import Lock
+from dataclasses import dataclass, field
 from datetime import UTC, datetime, timedelta
 from typing import cast
 
 import jwt
 from aiohttp import ClientSession, FormData
-from pydantic import BaseModel, Field, ValidationError
+from mashumaro import field_options
+from mashumaro.mixins.json import DataClassJSONMixin
```
WebSpider, line 17: Maybe use orjson instead?
https://github.com/Fatal1ty/mashumaro?tab=readme-ov-file#orjson-library

HA already depends on it and uses it for speed

myskoda/rest_api.py
```diff
@@ -38,7 +38,7 @@ async def get_info(self, vin: str) -> Info:
         ) as response:
             response_text = await response.text()
             _LOGGER.debug(f"vin {vin}: received basic info: {response_text}")
-            return Info.parse_raw(response_text)
+            return Info.from_json(response_text)
```
WebSpider, line 41: This will throw an exception if what we get back isnt valid JSON, which will happen on Skoda maintenance.
Consider adding exception handling code here

Prior99: I wrapped all get-calls in a `_make_get_request` method which performs the error handling

Prior99: See commit ad8d68a

myskoda/auth/authorization.py
Prior99: Done in f288bb8

WebSpider (approved):
Lgtm

PR #38, opened 2024-09-25
Optional user fields

dvx76:


dvx76:
Going to just merge and release this to unblock users but feel free to leave comments and I'll address them retroactively.

PR #40, opened 2024-09-26
Add missing CapabilityStatus

dvx76:
Fixes #34
Fixes skodaconnect/homeassistant-myskoda#35

PR #41, opened 2024-09-26
Fix missing field in ChargingStatus

dvx76:
Kinda weird this is missing only for some cars/users (when not charging)...

Fixes skodaconnect/homeassistant-myskoda#28


dvx76:
Lint errors fixed in #40

WebSpider (approved):
Lgtm

PR #42, opened 2024-09-26
Add calculated state of the doors and windows

fursov:
Calculated from renders that response messsage contains. Probably not very reliable method. Error cases assumed, return closed state if parsing fails for any reason.

Added tests for different states.

### DIFF
--- a/myskoda/models/status.py
+++ b/myskoda/models/status.py
@@ -1,12 +1,41 @@
 """Models for responses of api/v2/vehicle-status/{vin}."""
 
+import logging
 from dataclasses import dataclass, field
 from datetime import datetime
+from enum import IntEnum, StrEnum
+from urllib.parse import parse_qs, urlparse
 
 from mashumaro import field_options
 from mashumaro.mixins.orjson import DataClassORJSONMixin
 
 from myskoda.models.common import DoorLockedState, OnOffState, OpenState
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class CarBodyElements(IntEnum):
+    LEFT_FRONT_DOOR = 0
+    RIGHT_FRONT_DOOR = 1
+    LEFT_BACK_DOOR = 2
+    RIGHT_BACK_DOOR = 3
+    LEFT_LIGHT = 4
+    RIGHT_LIGHT = 5
+    TRUNK = 6
+    BONNET = 7
+    SUNROOF = 8
+
+
+class CarWindowDoorStates(IntEnum):
+    ALL_CLOSED = 1
+    WINDOW_OPENED = 2
+    DOOR_OPENED = 3
+
+
+class DoorWindowState(StrEnum):
+    CLOSED = "Closed"
+    DOOR_OPEN = "Door opened"
+    WINDOW_OPEN = "Window opened"
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


fursov:
I also would like to understand if this feature is ok to have. Please, provide your feedback.

myskoda/models/status.py
```diff
@@ -1,13 +1,36 @@
...
 
 from mashumaro import field_options
 from mashumaro.mixins.orjson import DataClassORJSONMixin
 
 from myskoda.models.common import DoorLockedState, OnOffState, OpenState
 
+_LOGGER = logging.getLogger(__name__)
+
+
+class CarBodyElements(IntEnum):
```
Prior99, line 17: We're naming enum members in `UPPER_SNAKE_CASE`.

myskoda/models/status.py
```diff
@@ -25,10 +48,83 @@ class Overall(DataClassORJSONMixin):
...
 @dataclass
 class Status(DataClassORJSONMixin):
     """Current status information for a vehicle."""
 
     car_captured_timestamp: datetime = field(metadata=field_options(alias="carCapturedTimestamp"))
     detail: Detail
     overall: Overall
+    renders: Renders
+
+    class _CarWindowDoorStates(IntEnum):
```
Prior99, line 74: Please, no classes inside of classes :)

myskoda/models/status.py
```diff
@@ -25,10 +48,83 @@ class Overall(DataClassORJSONMixin):
...
+        window_opened = 2
+        door_opened = 3
+
+    def _get_vehicle_state(self) -> list[int]:
+        try:
+            parsed_url = urlparse(self.renders.light_mode.one_x)
+            parsed_query = parse_qs(parsed_url.query)
+        except ValueError:
+            # Return default if parsing fails
+            _LOGGER.debug("Unable to parse vehicle status query for doors/windows state")
```
Prior99, line 85: This should log an error.

myskoda/models/status.py
```diff
@@ -25,10 +48,83 @@ class Overall(DataClassORJSONMixin):
...
     detail: Detail
     overall: Overall
+    renders: Renders
+
+    class _CarWindowDoorStates(IntEnum):
+        all_closed = 1
+        window_opened = 2
+        door_opened = 3
+
+    def _get_vehicle_state(self) -> list[int]:
```
Prior99, line 79: I think the method name is misleading.
This method takes the URL's and parses it into a list of integers.

`_extract_window_door_int_from_url`

Or something similar would be more fitting.

myskoda/models/status.py
```diff
@@ -25,10 +48,83 @@ class Overall(DataClassORJSONMixin):
...
+        try:
+            split_values = parsed_query["vehicleState"][0].split("-")
+        except (KeyError, IndexError):
+            split_values = ("1-1-1-1-0-0-0-0-0").split("-")
+        try:
+            integer_map = list(map(int, split_values))
+        except ValueError:
+            # Return default if mapping fails
+            _LOGGER.debug("Unable to deduct vehicle status for doors/windows state")
+            return [1, 1, 1, 1, 0, 0, 0, 0, 0]
```
Prior99, line 96: Instead of having multiple `try ... except` blocks which ultimately all lead to the method just returning `[1, 1, 1, 1, 0, 0, 0, 0, 0]`, you could just have one `try ... except` block around the whole body of the method and then return this here.
 
Will be less code and less computation.

myskoda/models/status.py
```diff
@@ -1,13 +1,36 @@
...
+    left_light = 4
+    right_light = 5
+    trunk = 6
+    bonnet = 7
+    sunroof = 8
+
+
+class DoorWindowState(StrEnum):
+    CLOSED = "Closed"
+    DOOR_OPENED = "Door opened"
```
Prior99, line 31: Please rename to `DOOR_OPEN` and `WINDOW_OPEN`.

tests/test_status.py
```diff
@@ -0,0 +1,117 @@
+"""Unit tests for myskoda.models.status."""
+
+from myskoda.models.status import DoorWindowState, Status
+
+FULL_CONFIG = {
```
Prior99, line 5: Please move this into a fixture file.

Prior99 (changes requested):
Awesome work. I was also planning to do this, but I never found the time.

Please take a look at the comments I've left you :)

myskoda/models/status.py
fursov: Done

myskoda/models/status.py
fursov: Done

myskoda/models/status.py
fursov: Done

myskoda/models/status.py
fursov: Done

myskoda/models/status.py
fursov: Done, return shorted list for doors/windows only, other elements states can be fetched from other Status fields

tests/test_status.py
fursov: Done

tests/test_status.py
```diff
@@ -0,0 +1,110 @@
...
+    vehicle_status = (FIXTURES_DIR / "superb/vehicle-status-doors-closed.json").read_text()
+    return json.loads(vehicle_status)
+
+
+def _get_config_with_lightmode_onex_url(full_config: dict, onex_url: str) -> dict:
+    full_config["renders"]["lightMode"]["oneX"] = onex_url
+    return full_config
+
+
+def test_door_window_states_all_closed_parsed_ok(vehicle_status: dict) -> None:
```
dvx76, line 25: No need to update the PR for this, just FYI, take a look at pytest parametrize (https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize-basics). It's awesome for tests like this where you're testing a various function inputs against expected output but structure of the test itself is identical between each test.

And thanks for including tests !

dvx76 (approved):

Prior99 (approved):

PR #47, opened 2024-09-27
(bug) Fix charging model to include errors

WebSpider:
When there is a charging error, the charging state is not always included in the API reply. Make the field optional
Also, include the charging error as we find them

Fixes https://github.com/skodaconnect/homeassistant-myskoda/issues/36

### DIFF
--- a/myskoda/models/charging.py
+++ b/myskoda/models/charging.py
@@ -89,7 +89,8 @@ class Battery(DataClassORJSONMixin):
 @dataclass
 class ChargingStatus(DataClassORJSONMixin):
     battery: Battery
-    state: ChargingState
+    state: ChargingState | None
+    errors: list[ChargingError] | None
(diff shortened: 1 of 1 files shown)


myskoda/models/charging.py
```diff
@@ -89,7 +89,8 @@ class Battery(DataClassORJSONMixin):
 @dataclass
 class ChargingStatus(DataClassORJSONMixin):
     battery: Battery
-    state: ChargingState
+    state: ChargingState | None
+    errors: list[ChargingError] | None
```
dvx76, line 93: Will these be `None/null`, or missing completely? In the later case I think we need to define a default value (of `None`) here, no?

WebSpider: They seem to be mutually exclusive:

- When there is an error, there is no status

I dont know for sure if there can be a state and an error at the same time

Prior99 (approved):

Issue #26, opened 2024-09-23
MQTT connection for CLI


WebSpider:
Let's discuss this in an issue and keep it out of this PR please :)

_Originally posted by @Prior99 in https://github.com/skodaconnect/myskoda/pull/10#discussion_r1771275581_

OBerghmans:
If I run the CLI I always get "Failed to connect to MQTT". I have not yet figured out whether it's something on my side, on the code side or just the MQTT server side.

However, I saw that quite some CLI commands don't really need the MQTT connection. Perhaps we can look into adjusting the CLI that it doesn't set up this connection if we're just going to make a API request and exit anyway.
That will prevent making unnecessary connections, but might make the CLI code a bit more complex.

Another option might be a "--disable-mqtt" flag, a bit more simple but less elegant.

Yesterday I commented out the mqtt part when preparing my PR as I didn't need the MQTT connection and the failure to connect prevented me from testing my changes.

OBerghmans:
An Enyaq 85x, this is the output in the CLI info command (mqtt lines were still commented out :-) )
```
model: Enyaq Coupé
model id: 5ACJK4
model year: 2025
title: Škoda Enyaq Coupé
```

OBerghmans:
The tcptraceroute works if I run it from a docker container. I just tried to run the CLI inside that container as well and that works too. Seems to be (my) windows related.

Anyway, apart from my issues, the goal of my comment was mainly the optimization of not running the MQTT if we don't need it. Like just now in my docker container, I ran the info command and it just connected and disconnected from MQTT without really needing it. Basically the same remark as @WebSpider made in that PR that led to this issue. 

We could add a flag on the myskoda class, with a default value of connecting to mqtt. From the CLI module we could then turn off the mqtt for commands that don't need it. That removes the unnecessary connection to mqtt and also all the related logs to this.
If you're open to the idea, I am willing to give it a shot and make a PR.

dvx76:
If it's just in the CLI that sounds good to me !

Prior99:
With the merge of #56 this is resolved.

PR #49, opened 2024-09-27
MQTT flag in CLI

OBerghmans:
An enable_mqtt flag was added to the MySkoda class. The flag is True by default so that MQTT will always be used unless we specify otherwise. 
The necessary checks are in place to set up (or not) the MQTT connection. A decorator that can be set on the methods that need MQTT will help us prevent long troubleshooting when we forget to enable the flag in the future. It would log an error if we disable MQTT but we try to do something with it anyway.

The CLI class was adjusted to do make the connection in the commands itself instead of the cli group. This is so that we can set MQTT disabled by default for CLI. A decorator helps us to easily indicate to enable MQTT. The ensure_connected function will retrieve all necessary information from the context and call the MySkoda connect method (with the correct setting of mqtt).

This PR relates to #26

### DIFF
--- a/myskoda/cli.py
+++ b/myskoda/cli.py
@@ -5,7 +5,7 @@ poetry run python3 -m myskoda.cli
 """
 
 import asyncio
-from collections.abc import Callable
+from collections.abc import Awaitable, Callable
 from functools import update_wrapper
 from logging import DEBUG, INFO
 from typing import ParamSpec, TypeVar
@@ -34,11 +34,13 @@ R = TypeVar("R")
 P = ParamSpec("P")
 
 
-def mqtt_required(func: Callable[P, R]) -> Callable[P, R]:
+def mqtt_required(func: Callable[..., Awaitable[R]]) -> Callable[..., Awaitable[Awaitable[R]]]:
+    """Enable MQTT before connecting to MySkoda."""
+
     @click.pass_context
-    def new_func(ctx: Context, *args: P.args, **kwargs: P.kwargs) -> R:
+    async def new_func(ctx: Context, *args, **kwargs) -> Awaitable[R]:  # noqa: ANN002, ANN003
         ctx.obj["myskoda"].enable_mqtt = True
-        return ctx.invoke(func, *args, **kwargs)
+        return await ctx.invoke(func, *args, **kwargs)
 
     return update_wrapper(new_func, func)
 
--- a/myskoda/cli.py
+++ b/myskoda/cli.py
@@ -5,7 +5,10 @@ poetry run python3 -m myskoda.cli
 """
 
 import asyncio
+from collections.abc import Callable
+from functools import update_wrapper
 from logging import DEBUG, INFO
+from typing import ParamSpec, TypeVar
 
 import asyncclick as click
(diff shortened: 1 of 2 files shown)


OBerghmans:
I'll take a look at the linting tomorrow. The pre-commit runs were finally OK with my current version, but the pipeline runs some other linting it seems.

WebSpider:
Thanks for your contribution!
I will add pyright to pre-commit as well, as we run this in the pipeline as well.

OBerghmans:
Ha finally! Now my head hurts of all the typing stuff. I hope it doesn't bother that I ignored the ANN002 and ANN003. Otherwise it became too complex.

myskoda/cli.py
```diff
@@ -45,8 +69,7 @@ async def cli(ctx: Context, username: str, password: str, verbose: bool) -> None
     if verbose:
         trace_configs.append(TRACE_CONFIG)
     session = ClientSession(trace_configs=trace_configs)
-    myskoda = MySkoda(session)
-    await myskoda.connect(username, password)
+    myskoda = MySkoda(session, enable_mqtt=False)
```
dvx76, line 72: I thought this was going to be behind a new `--mqtt` flag?

OBerghmans: You're correct that I said I would add that. I will put in the extra work to add that still.

Does your comment also mean that you prefer the flag only, or is the additional effort I've put in to not start the MQTT when it's not needed also nice?

OBerghmans: Also note that the `enable_mqtt=False` here does not mean that I disable it in general. Further down the code I turn it back on for the commands that need it (with the `@mqtt_required` decorator).

WebSpider: I think it would result in simpler code if mqtt connection is disabled by default for cli invocations, unless `--mqtt` is passed on the commandline.

Then you can keep mqtt connection the default for when the code is invoked from the integration

OBerghmans: The mqtt connection is still the default in the MySkoda class, as the default there is `True`. So I think we're fine for the integration using the mqtt connection by default. I see that the integration initiates the MySkoda class in the `__init__.py` and that would be with MQTT enabled.

For the CLI it is disabled by default. But I'm getting the idea that you prefer I remove the decorator and do the connection at the cli command group? What if the command needs MQTT but we don't give the --mqtt flag? I currently solved that with the decorator, as that enables mqtt whenever it's needed.

dvx76: I'll defer to @Prior99 here. As long as it's still on by default in the integration and can be toggled in the CLI (don't care too much what the default is) I'm good. I don't love `MySkoda` now requires a decorator on most current/future methods but I don't have a better proposal either 😅

Prior99: I'll have a look now

Prior99:
I only liked this partially, see my suggested implementation in #56

OBerghmans:
I was working on a new way as well, what about this?

Prior99:
Thank you for your contribution. I merged #56 instead, which contains some code I took from your PR.

PR #51, opened 2024-09-28
Fix seat_heating_activated being optional

dvx76:
I guess when the car doesn't have heated seats.

Fixes skodaconnect/homeassistant-myskoda#43


WebSpider (approved):
Lgtm

PR #68, opened 2024-09-30
Error handling for trip-statistics in CLI

### LINKED ISSUE #58: Trip statistics endpoint
I get the following error when running the CLI command `trip-statistics`:

```
2024-09-29 13:20:02 myskoda.myskoda[10496] INFO IDK Authorization was successful.
2024-09-29 13:20:02 myskoda.rest_api[10496] ERROR Failed to load data from url /v1/trip-statistics/TMBXYZ123?offsetType=week&offset=0&timezone=Europe%2FBerlin. Return value was 'Vehicle with VIN {TMBXYZ123} not found'
  File "C:\Users\oli\Local\personal\myskoda\myskoda\rest_api.py", line 41, in _make_get_request
    data = deserialize(response_text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 3, in __mashumaro_from_json__
orjson.JSONDecodeError: unexpected character: line 1 column 1 (char 0)
  File "<string>", line 1, in <module>
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\asyncclick\core.py", line 1205, in __call__
    return anyio.run(self._main, main, args, kwargs, **opts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\anyio\_core\_eventloop.py", line 74, in run
    return async_backend.run(func, args, {}, backend_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\anyio\_backends\_asyncio.py", line 2248, in run
    return runner.run(wrapper())
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\.pyenv\pyenv-win\versions\3.12.2\Lib\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\.pyenv\pyenv-win\versions\3.12.2\Lib\asyncio\base_events.py", line 685, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\anyio\_backends\_asyncio.py", line 2236, in wrapper
    return await func(*args)
           ^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\asyncclick\core.py", line 1208, in _main
    return await main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\asyncclick\core.py", line 1120, in main
    rv = await self.invoke(ctx)
         ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\asyncclick\core.py", line 1739, in invoke
    return await _process_result(await sub_ctx.command.invoke(sub_ctx))
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\asyncclick\core.py", line 1485, in invoke
    return await ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\AppData\Local\pypoetry\Cache\virtualenvs\myskoda-jII21Bdp-py3.12\Lib\site-packages\asyncclick\core.py", line 824, in invoke
    rv = await rv
         ^^^^^^^^
  File "C:\Users\oli\Local\personal\myskoda\myskoda\cli.py", line 316, in trip_statistics
    stats = await myskoda.get_trip_statistics(vin)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\Local\personal\myskoda\myskoda\myskoda.py", line 200, in get_trip_statistics
    return await self.rest_api.get_trip_statistics(vin)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\Local\personal\myskoda\myskoda\rest_api.py", line 83, in get_trip_statistics
    return await self._make_get_request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\oli\Local\personal\myskoda\myskoda\rest_api.py", line 41, in _make_get_request
    data = deserialize(response_text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 3, in __mashumaro_from_json__
orjson.JSONDecodeError: unexpected character: line 1 column 1 (char 0)
2024-09-29 13:20:02 asyncio[10496] ERROR Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x000001A98659FA70>
2024-09-29 13:20:02 asyncio[10496] ERROR Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x000001A98683B890>, 190083.359)]', '[(<aiohttp.client_proto.ResponseHandler object at 0x000001A986894AD0>, 190083.828)]']
connector: <aiohttp.connector.TCPConnector object at 0x000001A9868934A0>
```

Perhaps because my car is rather new and now trips were recorded yet?

OBerghmans:
I tried to add a minimal solution to better handle the trip-statistics error:
  - Added an exception in the rest_api module. The _make_get_request still throws an exception, so behaviour in the overall library should remain the same
  - The CLI catches the new exception and prints it

This should fix #58 , at least by printing the error result instead of throwing an exception.

It also takes into account @dvx76 's remark  that 

> The integration will use get_vehicle which will only try to get statistics if the TRIP_STATISTICS capability is present.

The CLI also still exposes all methods but does not error in case there are no trip statistics.

The fix is minimal in the sense that I think it would be overkill for the trip-statistics method to first check the capabilities and then - depending on the capability being present - get the trip statistics. But that's also a way we could go.

### DIFF
--- a/myskoda/cli.py
+++ b/myskoda/cli.py
@@ -15,6 +15,7 @@ import asyncclick as click
 import coloredlogs
 import yaml
 from aiohttp import ClientSession
+from aiohttp.client_exceptions import ClientResponseError
 from asyncclick.core import Context
 from pygments import highlight
 from pygments.formatters import TerminalFormatter
@@ -25,7 +26,6 @@ from myskoda.event import Event
 from myskoda.models.operation_request import OperationName
 
 from .myskoda import TRACE_CONFIG, MqttDisabledError, MySkoda
-from .rest_api import InvalidResponseError
 
 
 def mqtt_required[R](func: Callable[..., Awaitable[R]]) -> Callable[..., Awaitable[Awaitable[R]]]:
@@ -236,8 +236,8 @@ async def trip_statistics(ctx: Context, vin: str) -> None:
     try:
         stats = await myskoda.get_trip_statistics(vin)
         ctx.obj["print"](stats.to_dict())
-    except InvalidResponseError as e:
-        ctx.obj["print"]({"error": e.response})
+    except ClientResponseError as e:
+        ctx.obj["print"]({"error": e.status, "message": e.message, "url": str(e.request_info.url)})
 
 
 @cli.command()
--- a/myskoda/cli.py
+++ b/myskoda/cli.py
@@ -22,6 +22,7 @@ from termcolor import colored
 from .event import Event
 from .models.operation_request import OperationName
 from .myskoda import TRACE_CONFIG, MySkoda
+from .rest_api import InvalidResponseError
 
 
 class Format(StrEnum):
@@ -207,8 +208,11 @@ async def user(ctx: Context) -> None:
 async def trip_statistics(ctx: Context, vin: str) -> None:
     """Print the last trip statics."""
     myskoda: MySkoda = ctx.obj["myskoda"]
-    stats = await myskoda.get_trip_statistics(vin)
-    ctx.obj["print"](stats.to_dict())
+    try:
+        stats = await myskoda.get_trip_statistics(vin)
+        ctx.obj["print"](stats.to_dict())
+    except InvalidResponseError as e:
+        ctx.obj["print"]({"error": e.response})
 
 
 @cli.command()
--- a/myskoda/cli.py
+++ b/myskoda/cli.py
@@ -10,6 +10,7 @@ from collections.abc import Awaitable, Callable
 from enum import StrEnum
 from functools import update_wrapper
 from logging import DEBUG, INFO
+from typing import Any
 
 import asyncclick as click
 import coloredlogs
@@ -40,6 +41,18 @@ def mqtt_required[R](func: Callable[..., Awaitable[R]]) -> Callable[..., Awaitab
         return await ctx.invoke(func, *args, **kwargs)
 
     return update_wrapper(new_func, func)
+
+
+async def handle_request(ctx: Context, func: Callable[..., Awaitable], *args: Any) -> None:  # noqa: ANN401
+    """Handle API requests and perform error handling."""
+    try:
+        result = await func(*args)
+        if hasattr(result, "to_dict"):
+            ctx.obj["print"](result.to_dict())
+        else:
+            ctx.obj["print"](result)
+    except ClientResponseError as e:
+        ctx.obj["print"]({"error": e.status, "message": e.message, "url": str(e.request_info.url)})
 
 
 class Format(StrEnum):
@@ -133,9 +146,7 @@ async def auth(ctx: Context) -> None:
 @click.pass_context
 async def list_vehicles(ctx: Context) -> None:
     """Print a list of all vehicle identification numbers associated with the account."""
-    myskoda: MySkoda = ctx.obj["myskoda"]
-    vehicles = await myskoda.list_vehicle_vins()
-    ctx.obj["print"](vehicles)
+    await handle_request(ctx, ctx.obj["myskoda"].list_vehicle_vins)
 
 
 @cli.command()
@@ -143,9 +154,7 @@ async def list_vehicles(ctx: Context) -> None:
 @click.pass_context
 async def info(ctx: Context, vin: str) -> None:
     """Print info for the specified vin."""
-    myskoda: MySkoda = ctx.obj["myskoda"]
-    info = await myskoda.get_info(vin)
-    ctx.obj["print"](info.to_dict())
+    await handle_request(ctx, ctx.obj["myskoda"].get_info, vin)
 
 
 @cli.command()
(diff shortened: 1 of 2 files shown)


myskoda/cli.py
```diff
@@ -207,8 +208,11 @@ async def user(ctx: Context) -> None:
 async def trip_statistics(ctx: Context, vin: str) -> None:
     """Print the last trip statics."""
     myskoda: MySkoda = ctx.obj["myskoda"]
-    stats = await myskoda.get_trip_statistics(vin)
-    ctx.obj["print"](stats.to_dict())
+    try:
+        stats = await myskoda.get_trip_statistics(vin)
+        ctx.obj["print"](stats.to_dict())
+    except InvalidResponseError as e:
+        ctx.obj["print"]({"error": e.response})
```
Prior99, line 215: I like this, but can't this happen in every route? Not only just trip statistics. I think we can even add this in a generic way, and add this to the `_make_get_request`.

Prior99 (changes requested):

myskoda/cli.py
Prior99: Perhaps it would be even better to simply check the HTTP status code. Perhaps the response body will always be a string if it's not a 200 code?

OBerghmans: Checking the status code might be a good idea. I'll check what it is for the failed trip-statistics call.

For your first comment: a generic way can be checked. I think it would be mainly in the CLI class then. I'm not sure if we could add the generic way in the RestApi class, as we probably don't always want to just print that failure/response message. In a way I already did it generic: the entity using the _make_get_request (in this case the CLI) can decide what to do with the error. Perhaps sometimes we just want to catch it and not report on it.

OBerghmans: The status code was 404. I now use the raise_for_status built-in functionality for raising an exception if the result was not 200. The removes the need for a new exception (although I don't have the plain text response now). So let me know if we would rather have that one as well.

OBerghmans: I just pushed a more generic way in the CLI method. Closing this conversation on the outdated file. Looking forward to feedback though!

myskoda/rest_api.py
```diff
@@ -37,11 +36,10 @@ async def _make_get_request[T](self, url: str, deserialize: Callable[[str], T])
             f"{BASE_URL_SKODA}/api{url}",
             headers=await self._headers(),
         ) as response:
+            response.raise_for_status()
```
WebSpider, line 39: If we raise here, where do we catch the exception?

WebSpider (commented):
Small question, otherwise lgtm

myskoda/rest_api.py
OBerghmans: This will throw an exception (obviously) but in this situation the deserialize would have also threw an exception. So same result, just another exception class. I searched the code and everywhere else it was catched by 'except Exception' so this would still be the case. 

And in the CLI class I specifically catch this one to print a nicer output instead of exiting the app with a stack trace

OBerghmans: Also, the integration does extra checks (checking if the trip statistics capability is present before calling that endpoint) so that decreases the chances of getting the exception. There will probably be extra places where we can intercept this exception and improve. So in that point of view this is the start of some extra error handling when the API returns an error/Plain text/...

WebSpider (approved):
lgtm

PR #73, opened 2024-10-02
Fix air-conditioning and charging response processing

fursov:
Tests added.

Many fields have to be made optional because they do not exist in Superb responses.

I would appreciate if someone provides similar data from other cars so we can embed fixtures to the tests.


fursov:
Could someone suggest what the error in "Lint and test" means?

WebSpider:
It's the CI we run for every commit:
- Lint: Basic python linter, checks if there is syntax errors
- Test: Run coveralls, see if test coverage has improved

WebSpider (approved):
lgtm

Issue #79, opened 2024-10-05
AirConditioningState missing type "COOLING"


scoutice:
Traceback (most recent call last):
  File "<string>", line 31, in __mashumaro_from_json__
  File "C:\Program Files\Python312\Lib\enum.py", line 757, in __call__
    return cls.__new__(cls, value)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\enum.py", line 1171, in __new__
    raise ve_exc
ValueError: 'COOLING' is not a valid AirConditioningState

AirConditioning(timers=[Timer(enabled=False, id=1, time=datetime.time(7, 30), type=<TimerMode.ONE_OFF: 'ONE_OFF'>, selected_days=[<Weekday.WEDNESDAY: 'WEDNESDAY'>]), Timer(enabled=False, id=2, time=datetime.time(19, 59), type=<TimerMode.ONE_OFF: 'ONE_OFF'>, selected_days=[<Weekday.WEDNESDAY: 'WEDNESDAY'>])], car_captured_timestamp=datetime.datetime(2024, 10, 5, 12, 21, 47, tzinfo=datetime.timezone.utc), errors=[], state=<**AirConditioningState.COOLING: 'COOLING**'>, steering_wheel_position=<Side.LEFT: 'LEFT'>, window_heating_state=WindowHeatingState(front=<OnOffState.OFF: 'OFF'>, rear=<OnOffState.OFF: 'OFF'>, unspecified='INVALID'), air_conditioning_at_unlock=False, charger_connection_state=<ConnectionState.CONNECTED: 'CONNECTED'>, charger_lock_state=<ChargerLockedState.LOCKED: 'LOCKED'>, estimated_date_time_to_reach_target_temperature=datetime.datetime(2024, 10, 5, 12, 50, 52, 251836, tzinfo=datetime.timezone.utc), heater_source=None, seat_heating_activated=SeatHeating(front_left=True, front_right=False), target_temperature=TargetTemperature(temperature_value=15.5, unit_in_car=<TemperatureUnit.CELSIUS: 'CELSIUS'>), window_heating_enabled=True)

WebSpider:
Ack. This is an easy fix, accepting PR's until i have some time to fix it myself.

smiitm:
Hey, i have added 'COOLING' state to the AirConditioningState enum. This would be my first ever PR, plese let me know if i am making any mistake.

PR #80, opened 2024-10-05
#79 Add 'COOLING' state to AirConditioningState

smiitm:
This PR adds the missing 'COOLING' state to the AirConditioningState enum in response to issue #79. The issue reported a ValueError due to the missing 'COOLING' state, and I've updated the enum to include this state.

This is my first open source contribution, and I'm excited to be part of the community. Please feel free to provide any feedback or guidance if there's anything I can improve. Thanks.

### DIFF
--- a/myskoda/models/air_conditioning.py
+++ b/myskoda/models/air_conditioning.py
@@ -21,12 +21,13 @@ class TimerMode(StrEnum):
 
 
 class AirConditioningState(StrEnum):
+    COOLING = "COOLING"
     HEATING = "HEATING"
     HEATING_AUXILIARY = "HEATING_AUXILIARY"
     OFF = "OFF"
     ON = "ON"
     VENTILATION = "VENTILATION"
-    COOLING = "COOLING"
+
 
 # Probaly other states than AUTOMATIC are available, to be discovered
 class HeaterSource(StrEnum):
--- a/myskoda/models/air_conditioning.py
+++ b/myskoda/models/air_conditioning.py
@@ -26,7 +26,7 @@ class AirConditioningState(StrEnum):
     OFF = "OFF"
     ON = "ON"
     VENTILATION = "VENTILATION"
-
+    COOLING = "COOLING"
 
 # Probaly other states than AUTOMATIC are available, to be discovered
 class HeaterSource(StrEnum):


WebSpider:
Thanks for your contribution! 🙏
Please have a look at the linter results and correct your PR as needed.

myskoda/models/air_conditioning.py
```diff
@@ -26,7 +26,7 @@ class AirConditioningState(StrEnum):
     OFF = "OFF"
```
fursov, line 26: I also like when items are sorted alphabetically, it is easier to find the value and catch duplicates (may be not valid concern for such short enum but just to be consistent).

fursov:
Would be also nice to have a test case for this value. Please, have a look on `tests/test_rest_api.py` file for a function `load_air_conditioning()`. You can add your car's response to a fixture under folder `tests/fixtures` and then add the reference to the `load_air_conditioning()`

myskoda/models/air_conditioning.py
smiitm: Thanks for the suggestion! It seems practical, and I will definitely keep it in mind. I’ve made the change while fixing the other issue.

smiitm:
> Thanks for your contribution! 🙏 Please have a look at the linter results and correct your PR as needed.

hey, i've fixed the linting issue.

smiitm:
> Would be also nice to have a test case for this value. Please, have a look on `tests/test_rest_api.py` file for a function `load_air_conditioning()`. You can add your car's response to a fixture under folder `tests/fixtures` and then add the reference to the `load_air_conditioning()`

I will look into it. I wanted to confirm where the corresponding JSON file should be added. Should it go under enyaq/, superb/, or other/?

fursov:
> > Would be also nice to have a test case for this value. Please, have a look on `tests/test_rest_api.py` file for a function `load_air_conditioning()`. You can add your car's response to a fixture under folder `tests/fixtures` and then add the reference to the `load_air_conditioning()`
> 
> I will look into it. I wanted to confirm where the corresponding JSON file should be added. Should it go under enyaq/, superb/, or other/?

Depends on the car model for which you have your response. E.g. I have superb, so I add my jsons to superb folder. If you do not know the car model, put to other.

WebSpider (approved):
Lgtm

WebSpider:
Thanks for your contribution 👍

PR #81, opened 2024-10-06
Support case-insensitive StrEnums

### LINKED ISSUE #55: MQTT event field can have lowercase values
```
2024-09-29 10:20:13 WALLY myskoda.mqtt[25433] DEBUG Message (service-event) received for TMBxxx on topic charging: b'{"version":1,"traceId":"a1877638-cd2d-4907-8716-f9c0292e4794","timestamp":"2024-09-29T08:19:49Z","producer":"SKODA_MHUB","name":"change-soc","data":{"mode":"manual","state":"charging","soc":"67","chargedRange":"324","timeToFinish":"200","userId":"uuid","vin":"TMBxxxx"}}'
2024-09-29 10:20:13 WALLY asyncio[25433] ERROR Exception in callback Client.loop_read()
handle: <Handle Client.loop_read()>
  File "<string>", line 15, in __mashumaro_from_dict__
  File "/home/fdevaux/.pyenv/versions/3.12.2/lib/python3.12/enum.py", line 744, in __call__
    return cls.__new__(cls, value)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fdevaux/.pyenv/versions/3.12.2/lib/python3.12/enum.py", line 1158, in __new__
    raise ve_exc
ValueError: 'manual' is not a valid ChargeMode
```

- Option 1: Allow select Enums to ignore case (e.g. https://stackoverflow.com/a/76131490)
- Option 2: Deserialize all string values to lowercase and have all StrEnum values as lowercase. 

> Joostlek — Today at 10:42
> Lower case is nice because our translation files work with snake_case
> You could also consider making non essential values optional. In Withings I have a few enums that are optional, so i made a thing that puts in the log like, hey, this measurement unit is not part of the enum, please open an issue at this repo

dvx76:
Some endpoints and/or MQTT events return their values as uppercase for some users and lowercase for other users.

Fixes #55

### DIFF
--- a/myskoda/models/charging.py
+++ b/myskoda/models/charging.py
@@ -7,7 +7,7 @@ from enum import StrEnum
 from mashumaro import field_options
 from mashumaro.mixins.orjson import DataClassORJSONMixin
 
-from .common import ActiveState, EnabledState
+from .common import ActiveState, CaseInsensitiveStrEnum, EnabledState
 
 
 class ChargingErrorType(StrEnum):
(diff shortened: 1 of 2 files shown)


WebSpider (approved):
Lgtm

PR #83, opened 2024-10-06
Fix crashloop when failing to parse MQTT events

dvx76:
The same event gets processed over and over again if there is an exception in _on_message.

Not super happy with this fix but haven't found a better way. Discord discussion: https://discord.com/channels/877164727636230184/1087487796731707402/1292485603375448154

Fixes skodaconnect/homeassistant-myskoda#63


WebSpider (commented):
Are we expecting anything else than a JSONDecodeError?

dvx76:
> Are we expecting anything else than a JSONDecodeError?

`ValueError`, bunch of mashumaro exceptions like `mashumaro.exceptions.InvalidFieldValue`

WebSpider (approved):
Lgtm

PR #88, opened 2024-10-09
Add deserialization for time_to_finish

WebSpider:
Apparently, timeToFinish can be "null" according to issue skodaconnect/homeassistant-myskoda/issues/78

This adds a deserializer to handle the case

### DIFF
--- a/myskoda/models/service_event.py
+++ b/myskoda/models/service_event.py
@@ -89,10 +89,7 @@ def _deserialize_charging_state(value: str) -> ChargingState:
 def _deserialize_time_to_finish(value: int | str) -> int:
     if value == "null":
         return 0
-    try:
-        return int(value)
-    except (TypeError, ValueError) as err:
-        raise UnexpectedTimeToFinishError from err
+    return int(value)
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


dvx76 (dismissed):

myskoda/models/service_event.py
```diff
@@ -86,6 +86,12 @@ def _deserialize_charging_state(value: str) -> ChargingState:
             raise UnexpectedChargingStateError
 
 
+def _deserialize_time_to_finish(value: int | str) -> int:
+    if value == "null":
+        return 0
```
Prior99, line 91: Are you sure this is semantically a `0` and not rather `None`?

Prior99 (changes requested):

myskoda/models/service_event.py
WebSpider: The case here is that the car is currently not charging, so I guess a None would be more correct.
We would need to handle the case in the integration to determine what the None should correspond to.

I guess a sane state there would be "Unavailable"

Prior99: `None` corresponds to HA's unavailable.

dvx76 (approved):

PR #89, opened 2024-10-10
Add test for authorization

Prior99:


tests/fixtures/auth/signin.html
WebSpider, line 1: Do we want to use a fixture here, or actually grab the live contents from the web?
The latter would let us notice if they changed login (again)
Or maybe via a seperate test?

Prior99: Definitely the fixtures! Would we want to check-in credentials?
I don't feel comfortable having CI-tests run against live environment.
When it breaks, users will alert us quickly :D

dvx76 (approved):

PR #94, opened 2024-10-13
Improve fixture generator

Prior99:


dvx76 (dismissed):
Didn't review this too thoroughly but tested it out and it works and I'd rather just get this in and then update it as required vs. spending hours reviewing. Nice work!

dvx76 (approved):

PR #106, opened 2024-10-16
Add tests for MQTT and Operations

Prior99:
This PR adds tests for MQTT and all operations.

This also fixes some issues along the lines:

- We forgot to expose `set_charge_mode` as an operation
- We didn't wait until a topic was actually subscribed to
- Some aliases in the events from MQTT were wrong
- We didn't wait until the client is actually disconnected in `mqtt.disconnect`
- When using `honk_and_flash` with only `flash`, the wrong MQTT operation was awaited

And improves some other things:

- Events are now dataclasses
- CLI code is no longer taken into account for coverage report


WebSpider (approved):
LGTM!
Minor remark: Maybe we limit python to the version that HA supports as we ship the release. Right now this is 3.12 only.

PR #119, opened 2024-10-19
Anonymize VIN in errors saved in fixtures

### LINKED ISSUE #100: Errors not anonymized in fixtures
Generated fixtures still contain a vin in the errors.

dvx76:
Use regex to find VINs, based on VINs being 17 characters and the Skoda WMI being `TMB`.

Ref: https://www.skoda-storyboard.com/en/skoda-world/what-can-vin-codes-tell-you/

Fixes #100

In retrospect the original VIN is available where the additional call to `anonymize_url` is used so I could have kept the basic string `.replace`. But using regex allows this to be used from anywhere if we need to.

Also maybe I should rename `anonymize_url` to `anonymize_vin`?

### DIFF
--- a/myskoda/anonymize.py
+++ b/myskoda/anonymize.py
@@ -1,8 +1,11 @@
 """Methods for anonymizing data from the API."""
 
+import re
+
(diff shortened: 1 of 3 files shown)


WebSpider (changes requested):
Can we add a test so we know the regex is working?

WebSpider (approved):

Issue #123, opened 2024-10-22
ConcurrentSubscribe error upon reconnect


WebSpider:
myskoda version 0.8.4

Last entry in log on a normal instance running HA:
```
2024-10-22 13:05:55.199 DEBUG (MainThread) [custom_components.myskoda.coordinator] Finished fetching myskoda data in 3.261 seconds (success: True)
```

Suspend the device, and resume, and this happens:

```
2024-10-22 13:31:49.231 DEBUG (MainThread) [myskoda.mqtt] Socket to MQTT broker closed.
2024-10-22 13:31:49.231 DEBUG (MainThread) [myskoda.mqtt] Scheduling to reconnect MQTT.
2024-10-22 13:31:49.231 DEBUG (MainThread) [myskoda.mqtt] Connection to MQTT broker lost, reason 7.
2024-10-22 13:31:49.232 DEBUG (MainThread) [myskoda.mqtt] Scheduling to reconnect MQTT.
2024-10-22 13:31:49.232 DEBUG (MainThread) [myskoda.mqtt] Connecting to MQTT on mqtt.messagehub.de:8883...
2024-10-22 13:31:49.232 INFO (MainThread) [myskoda.auth.authorization] Token expired. Refreshing IDK access token
2024-10-22 13:31:49.737 DEBUG (MainThread) [myskoda.myskoda] Trace: POST https://mysmob.api.connect.skoda-auto.cz/api/v1/authenticati - response: 200 (4318 bytes) {"accessToken":"ey..4U","refreshToken":"ey..Bk","idToken":"ey..z4"}
2024-10-22 13:31:49.869 DEBUG (MainThread) [myskoda.mqtt] MQTT Connected.
2024-10-22 13:31:49.869 DEBUG (MainThread) [myskoda.mqtt] Subscribing to topic: 403cc5c0-d098-415d-8843-46ed171fd9b8/TMBJC9NY2MF017352/operation-request/air-conditioning/set-air-conditioning-at-unlock
2024-10-22 13:31:49.869 DEBUG (MainThread) [myskoda.mqtt] Subscribing to topic: 403cc5c0-d098-415d-8843-46ed171fd9b8/TMBJC9NY2MF017352/operation-request/air-conditioning/set-air-conditioning-at-unlock
  File "/home/me/.virtualenvs/homeassistant-skoda/lib/python3.12/site-packages/myskoda/mqtt.py", line 150, in connect
[...]

PR #129, opened 2024-10-27
Temporary workaround for MQTT ConcurrentSubscribeError and stability

dvx76:
Not fully clear about this yet but should at least help avoid skodaconnect/homeassistant-myskoda#140 and #123

When _something_ causes the MQTT connection to drop both the `on_disconnect` and the `on_socket_close` fire and both trigger a reconnect. This leads to a `ConcurrentSubscribeError` and all kinds of weird networking and stability issues throughout HA as a result.

### DIFF
--- a/myskoda/mqtt.py
+++ b/myskoda/mqtt.py
@@ -200,7 +200,6 @@ class Mqtt:
         self.is_connected = False
         if self._disconnect_listener is not None:
             self._disconnect_listener.set_result(None)
-        self.reconnect()
 
     def _on_connect_fail(self, client: AsyncioPahoClient, _data: None) -> None:
         if client is not self.client:


WebSpider (approved):
Lets see if this helps

WebSpider:
@dvx76 any idea what's causing the tests to fail?

dvx76:
> @dvx76 any idea what's causing the tests to fail?

Looking at it. I think there is at least one test which tests exactly this scenario where the socket gets closed abruptly without there being an on_disconnect event.

dvx76:
> > @dvx76 any idea what's causing the tests to fail?
> 
> Looking at it. I think there is at least one test which tests exactly this scenario where the socket gets closed abruptly without there being an on_disconnect event.

Alternative solution: keep the on_socket_close but remove the reconnect call in it. Need to test this out in HA but effect should be the same and it doesn't break the tests.

WebSpider (approved):
We probably will need to iterate on this to find the source, but let's go

PR #140, opened 2024-10-28
Replace asycnio-paho with aiomqtt

dvx76:
https://github.com/empicano/aiomqtt/tree/main

- asyncio-paho seems abandoned vs. aiomqtt actively developed
- seems simpler
- inspired by https://github.com/YoSmart-Inc/yolink-api/tree/main (used in HA integration)
- Don't pass in SSL Context to avoid impacting HA and other integrations (same as HA-core mqtt client)

---

Not done but seems to work fine, also in HA. Public methods remain the same (`connect()`, `subscribe()`, ...)

Current status:
- Seems to work fine with cli subscribe command as well as in HA
- Need to implement _running attribute as done in https://github.com/YoSmart-Inc/yolink-api/blob/main/yolink/mqtt_client.py#L56
- Struggling to get the unit test working. With some print-debugging I believe the problem is that the test will publish the test messages to the MQTT broker before the MySkodaMqttClient is actually connected and subscribed. Possibly caused by the connect and subscribe now running in an asyncio task ?


README.md
```diff
@@ -49,9 +49,6 @@ await myskoda.connect(email, password)
 
 for vin in await myskoda.list_vehicle_vins():
     print(vin)
-
-myskoda.disconnect()
-await session.close()
```
WebSpider, line 54: I do think adding in a disconnect is good hygiene :)

myskoda/mqtt.py
```diff
@@ -38,265 +39,212 @@
 _LOGGER = logging.getLogger(__name__)
 TOPIC_RE = re.compile("^(.*?)/(.*?)/(.*?)/(.*?)$")
 
+
+def _create_ssl_context() -> ssl.SSLContext:
+    """Create a SSL context for the MQTT connection."""
+    context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+    context.load_default_certs()
```
WebSpider, line 46: I think this is not asyncio safe, and will produce warnings with HA

myskoda/mqtt.py
```diff
@@ -38,265 +39,212 @@
...
+        self.hostname = hostname
         self.port = port
         self.enable_ssl = enable_ssl
+        self._callbacks = []
+        self._operation_listeners = []
+
+    async def connect(self, user_id: str, vehicle_vins: list[str]) -> None:
+        """Connect to the MQTT broker and listen for messages for the given user_id and VINs."""
+        _LOGGER.info("Connecting to MQTT with %s/%s", user_id, vehicle_vins)
+        print("XXXX Connecting to MQTT with %s/%s", user_id, vehicle_vins)
```
WebSpider, line 99: Debug statement present

myskoda/mqtt.py
```diff
@@ -38,265 +39,212 @@
...
-
-        self.is_connected = True
-
-        _LOGGER.debug("MQTT Connected.")
-        if not self.user_id or not self.vehicles:
-            _LOGGER.error("Reached on_connect, but user and vehicles not loaded")
+        Passing in pre-created SSLContext (vs 'tls_params=aiomqtt.TLSParameters()') to avoid a
+        blocking call in paho.mqtt.client. See https://github.com/w1ll1am23/pyeconet/pull/43.
+        """
+        _LOGGER.info("Starting _connect_and_listen")
```
WebSpider, line 128: Please change to debug

myskoda/mqtt.py
```diff
@@ -38,265 +39,212 @@
...
+        while True:
```
WebSpider, line 129: I think we should handle a single shot connection and listen in the private method. The loop we can implement in a public `connect_and_listen()`

myskoda/mqtt.py
```diff
@@ -38,265 +39,212 @@
...
+
+                    async for message in client.messages:
+                        print(f"XXXXXXX GOT MESSAGE {message}")
+                        self._on_message(message)
+            except aiomqtt.MqttError as exc:
+                _LOGGER.info("Connection lost (%s); reconnecting in %ss", exc, MQTT_RECONNECT_DELAY)
+                await asyncio.sleep(MQTT_RECONNECT_DELAY)
+
+    def _on_message(self, msg: aiomqtt.Message) -> None:
+        """Deserialize received MQTT message and emit Event to subscibed callbacks."""
```
WebSpider, line 160: Typo, subscribed

myskoda/mqtt.py
```diff
@@ -38,265 +39,212 @@
...
-
-    async def _subscribe_to_topic(self, topic: str) -> None:
-        _LOGGER.debug("Subscribing to topic: %s", topic)
-        if self._subscribed_listener is not None:
-            raise ConcurrentSubscribeError
-
-        future: Future[None] = get_event_loop().create_future()
-        self._subscribed_listener = future
+        _LOGGER.debug("Message (%s) received for %s on topic %s: %s", event_type, vin, topic, data)
+        print("XXX Message (%s) received for %s on topic %s: %s", event_type, vin, topic, data)
```
WebSpider, line 178: Debug statement present

myskoda/mqtt.py
WebSpider: Looked it up, this is safe

README.md
dvx76: I didn't mean to take the `session.close()` out but `MySkoda.disconnect()` has been removed in the PR. The only thing it was doing was disconnecting the mqtt client, which is now also gone because aiomqtt will call disconnect on the underlying paho mqtt client when exiting the context manager.

Also note that even in the current code nothing (other than 1 test) is actually calling `MySkoda.disconnect()`

myskoda/mqtt.py
dvx76: remove in next push

myskoda/mqtt.py
dvx76: in next push

myskoda/mqtt.py
dvx76: Can you provide more detail? I wanted to keep `connect()` as the public method for MySkoda to call. The function which contains this loop must be executed in an asyncio task. This was loosely based on https://github.com/YoSmart-Inc/yolink-api/blob/main/yolink/mqtt_client.py#L56 (lib used by a core HA integration of the same name) and the example at https://github.com/empicano/aiomqtt/blob/main/docs/reconnection.md

I should maybe also replicate the `while self._running:` bit and then we can have a disconnect method which cancels the asyncio task and sets _running to False (even if nothing is actually calling disconnect :) )

myskoda/mqtt.py
dvx76: fixed in next push

myskoda/mqtt.py
dvx76: fixed in next push

tests/test_mqtt.py
```diff
@@ -54,7 +84,7 @@ async def test_wait_for_operation(
 
 
 @pytest.mark.asyncio
-async def test_subscribe_event(
+async def no_test_subscribe_event(
```
dvx76, line 87: Ignore. Still trying to get tests to work.

tests/test_mqtt.py
```diff
@@ -36,6 +36,36 @@ async def test_wait_for_operation(
...
+    await mqtt_client.publish(
+        topic,
+        b'{"version":1,"operation":"start-air-conditioning","status":"IN_PROGRESS","traceId":"f0b638f7bba08ec4acb5de64cdb97ba9","requestId":"72f24950-b3db-4b7e-948f-7032f533773a"}',
+        QOS_2,
+    )
+    await mqtt_client.publish(
+        topic,
+        b'{"version":1,"operation":"start-air-conditioning","status":"IN_PROGRESS","traceId":"f0b638f7bba08ec4acb5de64cdb97ba9","requestId":"72f24950-b3db-4b7e-948f-7032f533773a"}',
+        QOS_2,
+    )
```
dvx76, line 68: This confirms my theory that the problem I have with the tests is that test messages get published before the new client is connected and subscribed. Repeating a bunch of publish calls like this the test starts to pass.

myskoda/mqtt.py
```diff
@@ -38,265 +40,221 @@
...
-
-        if not self.user_id or not self.vehicles:
-            raise NotConnectedError
-
-        task = create_task(cast(Any, self.connect(self.user_id, self.vehicles)))
-        background_tasks.add(task)
-        task.add_done_callback(background_tasks.discard)
+        self.vehicle_vins = vehicle_vins
+        self._listener_task = asyncio.create_task(self._connect_and_listen())
+        await self._subscribed.wait()
```
dvx76, line 106: Ok this is _a_ solution for my issue with the tests. I don't know if it's a good solution though 😅 

Basically forces a wait here until we've at least connected and subscribed once. This way when the test gets to the publish steps we're already subscribed and we get the test messages ...

WebSpider: I think it would be better to let the test-publisher wait until we should be able to connect

myskoda/mqtt.py
WebSpider: I see your point, but the `while True` just looks ugly :rofl: 

using `while self._running` as you suggested would be a nice improvement, so we have a way to stop this loop

pyproject.toml
```diff
@@ -11,9 +11,9 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 aiohttp = "^3"
+aiomqtt = "^2.0"
 asyncclick = { version = "^8.1.7.2", optional = true }
 asyncio = "^3"
-asyncio-paho = "^0.6.0"
 coloredlogs = { version = "^15.0.1", optional = true }
 paho-mqtt = "^1"
```
WebSpider, line 18: We can drop `paho-mqtt`, since `aiomqtt` takes care of that

myskoda/mqtt.py
dvx76: I would be less upset about this if it had a timeout so we don't block forever here if MQTT is unavailable (although I think that was the previous behavior).

I'm not sure how to make the test wait instead though. I guess I could make self.subscribed public and do the same `wait()` in the test?

myskoda/cli/__init__.py
```diff
@@ -119,8 +119,8 @@ async def disconnect(  # noqa: PLR0913
     myskoda: MySkoda = ctx.obj["myskoda"]
     session: ClientSession = ctx.obj["session"]
 
-    await myskoda.disconnect()
     await session.close()
+    await myskoda.disconnect()
```
dvx76, line 123: Need to swap the order back to original

myskoda/mqtt.py
dvx76: Implemented `self._running`

WebSpider (approved):
I've ran this for a few hours now and tried to break it in a few ways. Didn't work!

PR #161, opened 2024-11-06
Add files via upload

robertvanommen:
Skoda Octavia Combi 2023 fixture


WebSpider (approved):

WebSpider:
Thanks for this fixture!

We will add it to our list of cars that we test the library against, so any new features we add should become compatible with your car straight away.

PR #167, opened 2024-11-07
Add lock and unlock feature

prvakt:


dvx76:
Nice work! Can you take a look at `tests/test_operations.py` and see if you can add tests covering the new code there? Should be similar to existing operations.

myskoda/myskoda.py
```diff
@@ -217,6 +217,18 @@ async def stop_air_conditioning(self, vin: str) -> None:
         await self.rest_api.stop_air_conditioning(vin)
         await future
 
+    async def lock(self, vin: str, spin: str) -> None:
```
WebSpider, line 220: Is S-PIN a string? I would expect it to be an integer, but that is based on the input I can give in the app?

myskoda/rest_api.py
```diff
@@ -349,6 +349,24 @@ async def set_charge_mode(self, vin: str, mode: ChargeMode) -> None:
             json=json_data,
         )
 
+    async def lock(self, vin: str, spin: str) -> None:
+        """Lock the vehicle."""
+        _LOGGER.debug("Locking vehicle %s", vin)
+        json_data = {"currentSpin": spin}
```
WebSpider, line 355: Please add typing for the `json_data`

WebSpider (changes requested):
Nice work, thanks!

In addition to the adding of tests, please have a look at my small remarks.

myskoda/rest_api.py
```diff
@@ -349,6 +349,24 @@ async def set_charge_mode(self, vin: str, mode: ChargeMode) -> None:
             json=json_data,
         )
 
+    async def lock(self, vin: str, spin: str) -> None:
+        """Lock the vehicle."""
+        _LOGGER.debug("Locking vehicle %s", vin)
```
ThraaxSession, line 354: Prevent logging the VIN, it's a vey personal number and the logging wouldn't help here if there are issues. More importantly is the response, if the VIN or S-Pin is wrong.

myskoda/rest_api.py
```diff
@@ -349,6 +349,24 @@ async def set_charge_mode(self, vin: str, mode: ChargeMode) -> None:
...
+        json_data = {"currentSpin": spin}
+        await self._make_post_request(
+            url=f"/v1/vehicle-access/{vin}/lock",
+            json=json_data,
+        )
+
+    async def unlock(self, vin: str, spin: str) -> None:
+        """Unlock the vehicle."""
+        _LOGGER.debug("Unlocking vehicle %s", vin)
```
ThraaxSession, line 363: Same here.

myskoda/rest_api.py
prvakt: I just copied the complete function from another one and changed only name/parameters/endpoint, VIN is traced from every operation currently..

ThraaxSession: @WebSpider how do you think about it?

myskoda/myskoda.py
odelma: Can integers have leading zeros? In most cases I dont think so. And S-Pin can be e.g. 0000.

myskoda/rest_api.py
WebSpider: No problem with logging level debug. The integration supports multiple vehicles, so this helps in tracing code flow.

Dont log this in info and up.

WebSpider:
Tests were running for quite some time.

Please test locally by running 'just test' in the project's root (with `Just` installed)

myskoda/rest_api.py
WebSpider: I'm going to remove this remark, since we don't do this for other requests either.
When things quiet down, I'll refactor this all in one go.

myskoda/myskoda.py
dvx76: Yeah I think we want string here.

tests/test_operations.py
```diff
@@ -306,5 +306,47 @@ async def test_honk_and_flash(  # noqa: PLR0913
...
+@pytest.mark.parametrize("spin", ["1234", "4321"])
+async def test_unlock(
+    responses: aioresponses, mqtt_client: MQTTClient, myskoda: MySkoda, spin: str
+) -> None:
+    url = f"{BASE_URL_SKODA}/api/v1/vehicle-access/{VIN}/unlock"
+    responses.post(url=url)
+
+    future = myskoda.unlock(VIN, spin)
+
+    topic = f"{USER_ID}/{VIN}/operation-request/vehicle-access/lock-vehicle"
```
dvx76, line 343: Should be unlock I think?

```suggestion
    topic = f"{USER_ID}/{VIN}/operation-request/vehicle-access/unlock-vehicle"
```

dvx76 (approved):

WebSpider (approved):
Thanks for your contribution! 🖖

PR #187, opened 2024-11-09
More trip statistics optional

Cypherke:
These 2 fields need to be optional too. I guess this is because my car has been standing still at the dealers. I added my fixtures too to compare.


WebSpider (approved):
Yes, i figured as much when i did the last few conversion to optional.


Thanks for your contribution!

PR #226, opened 2024-11-18
combustion auxiliary-heater 

prvakt:
auxiliary-heater for combustion cars is using different request format so it needs to be handled separately


myskoda/myskoda.py
```diff
@@ -223,14 +223,18 @@ async def stop_air_conditioning(self, vin: str) -> None:
 
     async def start_auxiliary_heating(self, vin: str, temperature: float, spin: str) -> None:
         """Start the auxiliary heating with the provided target temperature in °C."""
-        # NOTE: 08/11/2024 - no response is published in MQTT (maybe bug in api?) so we don't wait
         future = self._wait_for_operation(OperationName.START_AUXILIARY_HEATING)
         await self.rest_api.start_auxiliary_heating(vin, temperature, spin)
         await future
 
+    async def start_combustion_auxiliary_heating(self, vin: str, timer: int, spin: str) -> None:
```
dvx76, line 230: Is there no capability or other field we can use to determine which endpoint we need to use (so that we could just have a single `start_auxiliary_heating` method)?

prvakt: yes there is a difference but request is different with different input field (temperature vs duration).

Hybrid is using `AIR_CONDITIONING_HEATING_SOURCE_AUXILIARY` capability with request structure:
```
{
    "airConditioningWithoutExternalPower": true,
    "heaterSource": "AUTOMATIC",
    "spin": "xxxx",
    "targetTemperature": {
        "temperatureValue": 21.5,
        "unitInCar": "CELSIUS"
    }
}
```
Combustion car  `AUXILIARY_HEATING` capability with request structure:
```
{
    "spin": "xxxx",
    "durationInSeconds": "600",
}
```

WebSpider: > Ah if they have a different signature (different arguments) it's less appealing to use a single method ☹️
> 
> @WebSpider what do you think?

Agree. Keep the methods simple and stack them if you want abstraction

WebSpider: > Capabilities are used only in HA to restrict specific entities . Should I rewrite it to use only single start_auxiliary_heating method and create request based on temperature or duration availability? 

It appears this is an example of a single entity that is valid for different capabilities but with a different method from the library to turn on/off. We should be able to do this. 

Instead of `required_capabilities` you could let `is_available()` do:

```python
def is_available(self) ->bool:
   return CapabilityId.EXAMPLE1 or CapabilityId.EXAMPLE2
```

and then change the methods in the entity to control what library call is done, depending on present capabilities or vehicle state (stored in `self.vehicle`)

prvakt: > It appears this is an example of a single entity that is valid for different capabilities but with a different method from the library to turn on/off. We should be able to do this.
> 
> Instead of `required_capabilities` you could let `is_available()` do:
> 
> ```python
> def is_available(self) ->bool:
>    return self.vehicle.has_capability(CapabilityId.EXAMPLE1) or self.vehicle.has_capability(CapabilityId.EXAMPLE2)
> ```
> 
> and then change the methods in the entity to control what library call is done for on/off/temperature, depending on present capabilities or vehicle state (stored in `self.vehicle`)

I think in HA we can use simple switch instead of a climate for cars with AUXILIARY_HEATING capability because heater can be only turned on or off - or you suggest to use climate entity in this case as well?

WebSpider: If we can only turn it on or off without any temperature settings or data, it should not be a climate entity

prvakt:
decided to have a look into decompiled MySkoda.apk and I think we need to rewrite it a little more
following parameters are allowed for start-auxiliary-heating request
```
      name = "spin"
      name = "durationInSeconds"
      name = "startMode"
      name = "targetTemperature"
```

- SPIN is mandatory, rest of the parameters are optional (or at least on my car, auxiliary heater will start with only SPIN request)
- durationInSeconds is probably available when AUXILIARY_HEATING_TIMERS are present
- startMode (probably HEATING or VENTILATION). Seems like HEATING is default and VENTILATION maybe if ACTIVE_VENTILATION is present ?
- targetTempereature is probably available when AUXILIARY_HEATING_TEMPERATURE_SETTING is present

WebSpider:
Thanks @prvakt for putting your time into this!

Looking at the complexity of your last remark, it may be a good idea to use some `model`s to work this out further, combined with 

1) A generic method in `myskoda` to handle the uniform call to the library which then
2) Splits up into multiple `myskoda.rest_api` calls for addressing the specific API calls down the line

Also: If you have any defaults, such as the `600` seconds for the default heater duration, please add them to `const.py` and import them, avoid using those directly in code.

prvakt:
> Thanks @prvakt for putting your time into this!
> 
> Looking at the complexity of your last remark, it may be a good idea to use some `model`s to work this out further, combined with
> 
> 1. A generic method in `myskoda` to handle the uniform call to the library which then
> 2. Splits up into multiple `myskoda.rest_api` calls for addressing the specific API calls down the line
> 
> Also: If you have any defaults, such as the `600` seconds for the default heater duration, please add them to `const.py` and import them, avoid using those directly in code.

But there is always only single API call to myskoda just with different parameters ... 

I don't know if I believe myself in such intervention in the code with my python knowledge especially If I don't have vehicle where I can test it before publishing to the world 🙈

prvakt:
tested everything & works as expected

myskoda/models/air_conditioning.py
```diff
@@ -36,6 +36,43 @@ class HeaterSource(StrEnum):
...
+
+        if self.source is not None:
+            json_data["heaterSource"] = self.source.value
+
+        if self.mode is not None:
+            json_data["startMode"] = self.mode.value
+
+        return json_data
+
+
```
WebSpider, line 75: This is redundant, it's already in the `DataClassORJSONMixin`: 
https://github.com/Fatal1ty/mashumaro?tab=readme-ov-file#json

WebSpider (changes requested):
The DataClassORJSONMixin already has a lot present for conversion to various formats. Please have a look at the mashumaru docs :+1:

prvakt:
seems like I can't overwrite to_json or to_dict but additional processing was needed so I have created get_config where this is done

when trying to serialize the data back to json or dictionary, it will return the field_name and seems like there is no easy way how to  use alias instead to build up needed request - I found a way but not sure if that's the proper way - if not please feel free to make changes by yourself because I have already lost half of my hair arguing with Chat GPT lol .
Also target temp needs to be rounded so that's additional processing which is also done in get_config

myskoda/models/air_conditioning.py
```diff
@@ -64,6 +72,38 @@ class WindowHeatingState(DataClassORJSONMixin):
...
+    )
+
+    class Config(BaseConfig):
+        """Configuration for serialization and deserialization.."""
+
+        code_generation_options = [  # noqa: RUF012
+            TO_DICT_ADD_BY_ALIAS_FLAG,
+            TO_DICT_ADD_OMIT_NONE_FLAG,
+        ]
+
```
WebSpider, line 97: @dvx76 requesting your input on this config

prvakt: this was only way how to ensure it will serialize dictionary with aliases and only those which are not none..

according to the mashumaro doc
`code_generation_options` config option
Some users may need functionality that wouldn't exist without extra cost such as valuable cpu time to execute additional instructions. Since not everyone needs such instructions, they can be enabled by a constant in the list, so the fastest basic behavior of the library will always remain by default. The following table provides a brief overview of all the available constants described below.

Constant	Description
[TO_DICT_ADD_OMIT_NONE_FLAG](https://pypi.org/project/mashumaro/#add-omit_none-keyword-argument)	Adds omit_none keyword-only argument to to_* methods.
[TO_DICT_ADD_BY_ALIAS_FLAG](https://pypi.org/project/mashumaro/#add-by_alias-keyword-argument)	Adds by_alias keyword-only argument to to_* methods.
[ADD_DIALECT_SUPPORT](https://pypi.org/project/mashumaro/#add-dialect-keyword-argument)	Adds dialect keyword-only argument to from_* and to_* methods.
[ADD_SERIALIZATION_CONTEXT](https://pypi.org/project/mashumaro/#add-context-keyword-argument)	Adds context keyword-only argument to to_* methods.

prvakt:
just to summarize my findings. There are 2 endpoints for retrieving air-conditioning related data
```
/v2/air-conditioning/{vin}
/v2/air-conditioning/{vin}/auxiliary-heating
```
PHEV vehicles (or at least my Superb iV) can retrieve AuxiliaryHeater status only from the first method 
ICE vehicles (based on the discussion and trace from discord) are using second endpoint to retrieve this data
#
PHEV vehicles can start auxiliary heating using both POST methods
```
/v2/air-conditioning/{vin}/start
/v2/air-conditioning/{vin}/auxiliary-heating/start
```

- if I put SPIN to first method and set `heaterSource`: `AUTOMATIC` car will start AUXILIARY heater (just don't know what will happen in summer when outside temperature will be bigger than  targetTemperature - I guess maybe the car will start cooling)

- for second method there is no need to set `heaterSource`, seems like its chosen automatically

For both methods I can also choose `targetTemperature`
#
ICE vehicles can start auxiliary heater using only second POST method
```
/v2/air-conditioning/{vin}/auxiliary-heating/start
```
additionally there are follwing options depending on the car

- `durationInSeconds`
- `targetTemperature`
- `startMode` (`VENTILATION` or `HEATING`)
#

I have implemented auxiliary-heater using the `/v2/air-conditioning/{vin}/auxiliary-heating` endpoint and with the possibility of pass all allowed parameters (source, duration, targetTemp and startMode). Only those which are **not None** will be serialized and sent inside the request

Retrieval of heating related data can be handled in HA by choosing `get_air_conditioning()` or `get_auxiliary_heating()` based on the car type / capability.

myskoda/models/air_conditioning.py
dvx76: Yep, I think this is the right way to do it. And better than just whipping up a dict of json_data locally within the rest_api method sending out a request, as we've been doing so far (e.g. `start_air_conditioning`)

💯

dvx76 (approved):
This is becoming quite large and has been open for a while so I don't want to scrutinize this too much and I'd rather just get it in so we can start using it in the integration and fix anything remaining later on.

But for what I have checked it all look solid to me! We should create models like this for the data we post to other endpoints as well!

If anything maybe the last comment where you summarize your findings so far is something worth moving into a markdown document in the docs directory so it doesn't get lost in a PR comment. But if you want to still do that it can be in a separate PR.

Thanks for all your work on this!

prvakt:
once I have whole air-conditioning (AC, Auxiliary Heater, Ventilation, Seats & Windows heating) fully implemented & working I will create a markdown document with all my findings and knowledge ;)

WebSpider:
This gave a small incompatibility in a climate entity in the `homeassistant-myskoda` component, which was fixed by commit https://github.com/skodaconnect/homeassistant-myskoda/pull/349/commits/0df7796eea82c3a8c697735ae0c9700a5200e028

Note to all involved and self: 
Next time, please make a clear mark in the PR when you alter method invocation. I must admit I missed this as well while reviewing.

PR #236, opened 2024-11-24
Implement 3 missing AC setting methods

prvakt:
- set_ac_at_unlock (enable/disable AC at unlock)	
- set_seats_heating (enable/disable seats[frontLeft,frontRight] heating with AC)
- set_windows_heating (enable/disable windows heating with AC)

### DIFF
--- a/myskoda/rest_api.py
+++ b/myskoda/rest_api.py
@@ -29,7 +29,13 @@ from myskoda.models.position import Position, PositionType
 
 from .auth.authorization import Authorization
 from .const import BASE_URL_SKODA, REQUEST_TIMEOUT_IN_SECONDS
-from .models.air_conditioning import AirConditioning, SeatHeating
+from .models.air_conditioning import (
+    AirConditioning,
+    AirConditioningAtUnlock,
+    AirConditioningWithoutExternalPower,
+    SeatHeating,
+    WindowHeating,
+)
 from .models.auxiliary_heating import AuxiliaryConfig, AuxiliaryHeating
 from .models.charging import Charging
 from .models.driving_range import DrivingRange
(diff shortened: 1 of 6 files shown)


myskoda/rest_api.py
```diff
@@ -327,6 +327,33 @@ async def set_ac_without_external_power(self, vin: str, enabled: bool) -> None:
...
+        json_data = {"airConditioningAtUnlockEnabled": "True" if enabled else "False"}
+        await self._make_post_request(
+            url=f"/v2/air-conditioning/{vin}/settings/ac-at-unlock",
+            json=json_data,
+        )
+
+    async def set_windows_heating(self, vin: str, enabled: bool) -> None:
+        """Enable or disable windows heating with AC."""
+        _LOGGER.debug("Setting windows heating with AC for vehicle %s to %r", vin, enabled)
+        json_data = {"windowHeatingEnabled": "True" if enabled else "False"}
```
WebSpider, line 342: Why does the seat-heating message have a proper class, but window-heating doesnt?

myskoda/rest_api.py
```diff
@@ -327,6 +327,33 @@ async def set_ac_without_external_power(self, vin: str, enabled: bool) -> None:
             json=json_data,
         )
 
+    async def set_ac_at_unlock(self, vin: str, enabled: bool) -> None:
+        """Enable or disable AC at unlock."""
+        _LOGGER.debug("Setting AC at at unlock for vehicle %s to %r", vin, enabled)
+        json_data = {"airConditioningAtUnlockEnabled": "True" if enabled else "False"}
```
WebSpider, line 333: Why does the seat-heating message have a proper class, but ac-unlock doesnt?

WebSpider (commented):
Thanks for this awesome addition! 👍

Some questions about consistency, but this looks like solid code 👍

prvakt:
> Thanks for this awesome addition! 👍
> 
> Some questions about consistency, but this looks like solid code 👍

request for seats heating can contain one or more seats so its needs to be handled in more special way (like the auxiliary heater), but windows_heating or ac_at_unlock can be only activated or deactivated using always same parameter so I did not wanted to complicate it too much.. But of course I can modify it so its handled same as seats_heating

WebSpider:
I do prefer using the same structure, even if it looks like some overhead. Python will optimize it, and it produces more predictable code

PR #240, opened 2024-11-24
added missing Endpoint.AUXILIARY_HEATING and refactored

prvakt:


WebSpider:
Have you tested this?

prvakt:
> Have you tested this?

I just forgot to add `Endpoint.AUXILIARY_HEATING` in the get_endpoint() method when implementing Auxiliary Heating in my previous PR.. For `get all` it was already trying to retrieve data from this endpoint but like I wrote already before, Superb iV doesn't use this endpoint but air-conditioning is used instead.. From discord I got some sample response so I have tested it at least with that

prvakt:
> Have you tested this?

and yes I have tested get specific fixture using the `get "keyword"` option and it work as before, there is just much less `if` conditions in the code ;)

dvx76 (approved):

WebSpider:
Thanks for your contributions! 🍻

PR #260, opened 2024-12-03
Fix Auxiliary Heater for ICE, add outside_temperature

prvakt:
There is a difference between how state from auxiliary heater can be obtained between ICE and PHEV
- PHEV cars are not using auxiliary-heating endpoint at all and state is available via air-conditioning
- ICE cars are using both endpoints, but state inside air-conditioning is always `state="INVALID"` independent on the real state. Because of that, we must retrieve state from auxiliary-heating endpoint. Because of that, auxiliary-heating data have been added to the Vehicle class and can be retrieved via get_vehicle()

+ added outside_temperature as it can be retrieved from auxiliary-heating endpoint if car with AUXILIARY_HEATING and OUTSIDE_TEMPERATURE capability is present
+ added AUXILIARY_HEATING_BASIC capability


myskoda/myskoda.py
```diff
@@ -354,6 +354,9 @@ async def get_vehicle(self, vin: str) -> Vehicle:
         if info.is_capability_available(CapabilityId.AIR_CONDITIONING):
             vehicle.air_conditioning = await self.get_air_conditioning(vin)
 
+        if info.is_capability_available(CapabilityId.AUXILIARY_HEATING):
+            vehicle.auxiliary_heating = await self.get_auxiliary_heating(vin)
+
```
WebSpider, line 359: Do we need to poll if we have `AUXILIARY_HEATING_BASIC` as well?

WebSpider (commented):
Thanks for your continued work on the heater! :muscle:

myskoda/myskoda.py
prvakt: Yes.. after reviewing other fixtures I must say its AUXILIARY_HEATING.. All cars with this capability from our fixtures have `state="INVALID"` in air-conditioning state...  `AUXILIARY_HEATING_BASIC` is a new capability which was present on a single car (new Superb) and so far looks its not usable anywhere, but just wanted to have it in the enum

WebSpider (approved):

PR #270, opened 2024-12-05
Fix charging-completed event

fursov:
Seems, in different versions of the car SW the event might contain or might not contain additional data. I can provide additional logs later (from Superb iV 2020), currently those are vanished after few reboots of the HA.

### DIFF
--- a/myskoda/models/service_event.py
+++ b/myskoda/models/service_event.py
@@ -161,7 +161,11 @@ class ServiceEventChargingCompleted(ServiceEvent):
     """Event class for charging-completed service event."""
 
     name = ServiceEventName.CHARGING_COMPLETED
-    data: ServiceEventChargingData
+    # ServiceEventChargingData should be given first in the union list
+    # because the ServiceEventData is a subset of the ServiceEventChargingData
+    # and will be recognized as valid parser but then it will miss the rest of
+    # the data
+    data: ServiceEventChargingData | ServiceEventData
 
 
 class ServiceEventChargingStatusChanged(ServiceEvent):
--- a/tests/test_mqtt.py
+++ b/tests/test_mqtt.py
@@ -158,6 +158,22 @@ async def test_subscribe_event(
                 }
             ),
         ),
+        (
+            f"{base_topic}/service-event/charging",
+            json.dumps(
+                {
+                    "version": 1,
+                    "traceId": trace_id,
+                    "timestamp": timestamp_str,
+                    "producer": "SKODA_MHUB",
+                    "name": "charging-completed",
+                    "data": {
+                        "userId": USER_ID,
+                        "vin": VIN,
+                    },
+                }
+            ),
+        ),
     ]
 
     def on_event(event: Event) -> None:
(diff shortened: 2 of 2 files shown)


myskoda/models/service_event.py
```diff
@@ -161,7 +161,11 @@ class ServiceEventChargingCompleted(ServiceEvent):
     """Event class for charging-completed service event."""
 
     name = ServiceEventName.CHARGING_COMPLETED
-    data: ServiceEventChargingData
+    # ServiceEventChargingData should be given first in the union list
+    # because the ServiceEventData is a subset of the ServiceEventChargingData
+    # and will be recognized as valid parser but then it will miss the rest of
+    # the data
+    data: ServiceEventChargingData | ServiceEventData
```
WebSpider, line 168: If we accept `ServiceEventData` here, we should make sure we handle the change in any part of the lib or integration that expects the additional fields of `ServiceEventChargingData`

fursov: We do not process the event yet anywhere. I was planning to add it as part of https://github.com/skodaconnect/homeassistant-myskoda/issues/370

WebSpider: The issue is that everywhere we accept `ServiceEventChargingData` now (for instance to handle updated SoC) , we should also make sure to handle what happens if we end up with `ServiceEventData` instead.

fursov: @dvx76 did you mean something like this? https://github.com/skodaconnect/myskoda/compare/main...fursov:myskoda:fix-charging-stopped-event-2

I did not place this change as pull request here, just keep it for discussion in my fork.

fursov: @WebSpider @dvx76 Please, consider the solutions I provided. Either this one, which will work well because https://github.com/skodaconnect/homeassistant-myskoda/pull/417 has been merged. Or review the one I mentioned just above.
Or let me know if none of it makes sense. We can discuss some better solution. The original problem is still there, so in my case HA integration does not receive `charging-completed` event.

WebSpider: > @dvx76 did you mean something like this? https://github.com/skodaconnect/myskoda/compare/main...fursov:myskoda:fix-charging-stopped-event-2
> 
> I did not place this change as pull request here, just keep it for discussion in my fork.

@dvx76 could you see if this is more in line with what you mean?

dvx76: Very sorry for the delay in responding here. Yes this is what I had in mind although I was hoping we could be even more deterministic than that and we would know exactly when we get either a `ServiceEvent` or a `ServiceEventWithChargingData`. But perhaps that's not really possible. I don't have time to look deeply into it and I won't want to block this further so I'm fine the proposed change and we can always come back to it later if needed.

fursov: #308 is opened for this

dvx76: Cool. So this can be closed then?

fursov: yes

PR #283, opened 2024-12-14
add missing state for both Window and Door open

prvakt:
working on implementation of binary sensor for each door to detect window/door opened/closed and noticed we are missing state where both door and window is opened (state=4)

### DIFF
--- a/myskoda/models/status.py
+++ b/myskoda/models/status.py
@@ -30,12 +30,14 @@ class CarWindowDoorStates(IntEnum):
     ALL_CLOSED = 1
     WINDOW_OPEN = 2
     DOOR_OPEN = 3
+    ALL_OPEN = 4
 
 
 class DoorWindowState(StrEnum):
     CLOSED = "Closed"
     DOOR_OPEN = "Door opened"
     WINDOW_OPEN = "Window opened"
+    ALL_OPEN = "Door and Window opened"
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


WebSpider (approved):
Lgtm, but wonder what we use DoorWindowState for. If we use lower_snake_case, it would fit HA translations better

prvakt:
> Lgtm, but wonder what we use DoorWindowState for. If we use lower_snake_case, it would fit HA translations better

we are using just key from DoorWindowState, not the string, but maybe we can remove DoorWindowState and keep only CarWindowDoorStates?

WebSpider:
Yes, agree. Feel free to mark it as deprecated, removal in v1.0.0

prvakt:
ok, will update tests accordingly to use 

> Yes, agree. Feel free to mark it as deprecated, removal in v1.0.0
> 
> According to semver, we only make breaking changes in major versions. Removal of a class is breaking.
> 
> No idea yet when v1.0.0 will happen, not in 2024😄

Even if the class is used just internally in one private method? I have removed CarWindowDoorStates because its not used anywhere, only by private `_extract_window_door_state_list_from_url()` and changed the DoorWindowState from StrEnum to IntEnum... We are not using those classes anywhere in the code yet (just DoorWindowState in tests)

WebSpider:
If it is not exposed in any way, there is no problem indeed

myskoda/models/status.py
```diff
@@ -26,16 +26,12 @@ class CarBodyElements(IntEnum):
...
     WINDOW_OPEN = 2
     DOOR_OPEN = 3
-
-
-class DoorWindowState(StrEnum):
-    CLOSED = "Closed"
-    DOOR_OPEN = "Door opened"
-    WINDOW_OPEN = "Window opened"
+    ALL_OPEN = 4
+    UNKNOWN = 0  # default state for invalid values
```
WebSpider, line 34: By introducing a default and mapping unknowns to it, we will miss new states, at the profit of causing a serialisation issue when there are new values.

@dvx76 do you have any preference?
I would prefer to keep an exception in the library and handle that gracefully in the integration, since the library is about determined state, not assumed state.

prvakt: before I made those changes, all errors were mapped to value 1 - ALL_CLOSED which was in my opinion not correct

WebSpider (approved):

Issue #277, opened 2024-12-10
Redirect to marketing consent after auth


Thor77:
I encountered a problem with my Skoda ID (in germany) that required me to debug the authorization code to find the problem:
There was a redirect to marketing consent in place after login, could be reproduced by manually logging in to https://skodaid.vwgroup.io/account
The redirect location looked something like this: `https://identity.vwgroup.io/signin-service/v1/consent/marketing/[...]`
Would you accept a PR to add a special case in https://github.com/skodaconnect/myskoda/blob/ee14a3b5ba2f243794e19dfab88b8f097c5da1d4/myskoda/auth/authorization.py#L163 so a more meaningful error is returned instead of a KeyError because the Location header is missing?
This could be implemented the same way the TermsAndConditionsException is implemented.

dvx76:
A PR would be great 👍 

See also https://github.com/skodaconnect/homeassistant-myskoda/issues/372

Thor77:
I created #284

WebSpider:
This has been merged and will appear in the next release. Thanks for your contribution! :partying_face:

PR #284, opened 2024-12-15
Add exception for redirect to marketing consent

Thor77:
during login the process can halt because the user is redirected to a marketing constent page. Without catching this condition a KeyError Location is raised because it's the final page but doesn't start with `myskoda://`.

Ref #277

### DIFF
--- a/myskoda/auth/authorization.py
+++ b/myskoda/auth/authorization.py
@@ -164,7 +164,7 @@ class Authorization:
                     if "terms-and-conditions" in location:
                         raise TermsAndConditionsError(location)
                     if "consent/marketing" in location:
-                        raise MarketingConstentError(location)
+                        raise MarketingConsentError(location)
                     async with self.session.get(location, allow_redirects=False) as response:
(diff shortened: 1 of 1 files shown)


myskoda/auth/authorization.py
```diff
@@ -163,6 +163,8 @@ async def _enter_password(self, csrf: CSRFState) -> IDKAuthorizationCode:
                 while not location.startswith("myskoda://"):
                     if "terms-and-conditions" in location:
                         raise TermsAndConditionsError(location)
+                    if "consent/marketing" in location:
```
WebSpider, line 166: Please rename exception to MarketingConsentError, it contains a typo now

WebSpider (changes requested):

myskoda/auth/authorization.py
Thor77: argh, sry, fixed

WebSpider (approved):
lgtm

PR #287, opened 2024-12-16
Update nx53l5_octavia_2024_combi

reharmsen:
Generated new fixture with latest MySkoda CLI (0.15.4) to update old fixture generated with 0.8.3

Does give an exception at the auxiliary_heating endpoint though


WebSpider:
Thanks!

Regarding the exception: Does your car have an auxiliary heating system?

WebSpider (approved):

PR #288, opened 2024-12-17
add departure info feature

prvakt:
- Add get DepartureInfo
- Add set DepartureTimer
- Add CLI options to for get DepartureInfo and set DepartureTimer
- Extend get_partial_vehicle and generate_fixtures with DepartureInfo
- Added tests

### DIFF
--- a/myskoda/models/departure.py
+++ b/myskoda/models/departure.py
@@ -1,0 +1,109 @@
+"""Models for responses of api/v1/vehicle-automatization/{vin}/departure/ endpoint."""
+
+from dataclasses import dataclass, field
+from datetime import datetime
+from datetime import time as datetime_time
+from typing import Any
+
+from mashumaro import field_options
+from mashumaro.config import BaseConfig
+from mashumaro.mixins.orjson import DataClassORJSONMixin
+
+from .air_conditioning import TemperatureUnit, TimerMode
+from .common import Weekday
+
+
+@dataclass
+class DepartureTemperature(DataClassORJSONMixin):
+    celsius: float | None = field(default=None, metadata=field_options(alias="celsius"))
+    fahrenheit: float | None = field(default=None, metadata=field_options(alias="fahrenheit"))
+    unit_in_car: TemperatureUnit = field(
+        default=TemperatureUnit.CELSIUS, metadata=field_options(alias="unitInCar")
+    )
+
+
+@dataclass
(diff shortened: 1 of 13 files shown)


myskoda/models/departure.py
```diff
@@ -0,0 +1,109 @@
...
+class DepartureTemperature(DataClassORJSONMixin):
+    celsius: float | None = field(default=None, metadata=field_options(alias="celsius"))
+    fahrenheit: float | None = field(default=None, metadata=field_options(alias="fahrenheit"))
+    unit_in_car: TemperatureUnit = field(
+        default=TemperatureUnit.CELSIUS, metadata=field_options(alias="unitInCar")
+    )
+
+
+@dataclass
+class ChargingTime(DataClassORJSONMixin):
```
WebSpider, line 26: Why use ChargingTime here?

WebSpider (commented):
Wow, you are on fire! 💥

I am mobile right now, so will look in detail later. Found one class that I dont understand why it is in this PR at first glance.

myskoda/models/departure.py
prvakt: its part of the DepartureInfo response and also required in set DepartureTimer request
You can specify time range when you want your car to be charged automatically so car will start/stop charging or air conditioning by itself if timer is active and conditions from the departure timer are met.

WebSpider (changes requested):

myskoda/rest_api.py
```diff
@@ -514,6 +530,21 @@ async def flash(
...
+    async def set_departure_timer(self, vin: str, timer: DepartureTimer) -> None:
+        """Set departure timer."""
+        _LOGGER.debug(
+            "Setting departure timer #%i for vehicle %s to %r", timer.id, vin, timer.enabled
+        )
+
+        now = datetime.now(UTC)
+        datetime_str = now.isoformat()
+
+        json_data = {"deviceDateTime": datetime_str, "timers": [timer.to_dict()]}
```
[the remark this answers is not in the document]
prvakt: to be honest I don't know - but API will return error in case its not send even it's not stored anywhere. MySkoda app is sending current timestamp as well.

WebSpider (approved):
Great stuff 👍

Dont forget to add the new topic / api usage in documentation

PR #290, opened 2024-12-17
add CableLock feature (set_auto_unlock_plug)

prvakt:
Enable or disable auto unlock plug when charged
(needs to be tested as I don't have this feature and just took it from decompiled apk)


WebSpider (approved):
lgtm, thanks!

PR #293, opened 2024-12-19
Add set-minimum-charge-limit feature

prvakt:
Part of departure timers - used when there is at least 1 active timer. Car will charge to the specified SoC after connecting charging cable and stop charging when specified SoC is reached. Rest is charged using departure timer rules.


WebSpider (commented):
Do we have any idea of the errors returned when we pass an impossible parameter, such as - 10,105 or when there is no timer?
If so, it would be great if we could make exceptions for them straight away

WebSpider (approved):
otherwise, lgtm

PR #302, opened 2024-12-28
send current timestamp inside departure info request

prvakt:
this will possibly fix  the problem with citigo responding with error 500 from departure-info endpoint. MySkoda app is also sending current timestamp in the url so we should do the same.. Just seems like API for some cars (for example Superb iV) is not checking if this is available or not and returning departure info data anyway


myskoda/rest_api.py
```diff
@@ -283,7 +284,20 @@ async def get_departure_timers(
         self, vin: str, anonymize: bool = False
     ) -> GetEndpointResult[DepartureInfo]:
         """Retrieve departure timers for the vehicle."""
-        url = f"/v1/vehicle-automatization/{vin}/departure/timers"
+        # Get the current local time with timezone
+        now = datetime.now().astimezone()
+        # Format the datetime string with timezone
+        formatted_time = (
```
WebSpider, line 290: Is there no simpler way to achieve the desired formatting?

WebSpider (approved):
Minor formatting question, otherwise, lets try it

myskoda/rest_api.py
prvakt: unfortunately not, because `%z` is without colon(`+0100`), but API is expecting timezone with : between hours and minutes..

And strftime function in Python does not include a built-in option to format the timezone offset with a colon directly

PR #303, opened 2024-12-28
Change MQTT connection to behave more like MySkoda app

prvakt:
couple of changes to mimic MySkoda app behaviour
- MQTT_KEEPALIVE = 60
- clean_session = true
- generate UUID only once and keep same as long as the application is running
- first 10 mqtt reconnects without backoff. MySkoda app is connecting to MQTT usually 2-5 times until its succeeded


WebSpider (approved):
Lgtm

PR #305, opened 2024-12-31
Revert "send current timestamp inside departure info request"

prvakt:
Reverts skodaconnect/myskoda#302

one of the tests is failing, already working on a fix


prvakt:
fixed by PR #306

PR #308, opened 2025-01-02
Refactor service event to distinguish events with data or without

fursov:
Only charging events have data and some of them have the data optionally. It could make sense to get only two types of events: with data or without. Currently only charging events will have these types, all other service events will be of ServiceEvent base type.

This solution has pros and cons:
- this is simpler and shorter
- only two event types to distinguish and process But:
- this solution is not future proof: if any of the event start to get other data, it would be required to introduce new types anyway which will make mqtt._parse_topic function to look even more ugly and complicated than now.

**NOTE!!!** This change is _probably_ not compatible with current one that is used in HA addon. Interface adaptation _might be_ required.


WebSpider:
Just thinking out loud here: Would it make sense to have 3 classes: `ServiceEventWithoutData`, `ServiceEventWithData`, which then subclasses into `ServiceEventWithChargingData`?

That way, we're flexible for future updates to service events that contain data

fursov:
@WebSpider I started to redesign this with your proposal but came up to conclusion that I would probably start to make too many assumptions about potential future requirements. I would propose that I keep the change as is now. If we start to get other `data` in the future, it will be a lot more clear which direction to go. Then we will refactor that code to match the needs.

WebSpider:
Makes sense. I will review the changes once we have a stable release again.

fursov:
Is there way to test this change with HA integration? I'm not 100% sure if it is fully API compatible.
One way is to have the library release without using it right away in HA integration release and then I can test it locally and make HA integration modifications if needed.

WebSpider:
> Is there way to test this change with HA integration? I'm not 100% sure if it is fully API compatible. One way is to have the library release without using it right away in HA integration release and then I can test it locally and make HA integration modifications if needed.

I usually have a local version of the library installed in a dev HA install, but if it makes things easier for you, we can do a pre-release of the lib, so you can test it.

fursov:
yes, pre-release would be better for me, I do not have dev HA installation atm.

dvx76 (approved):

WebSpider:
https://github.com/skodaconnect/myskoda/releases/tag/v0.17.0%2Beventdata

fursov:
I'll test this today-tomorrow.

WebSpider:
Had to rename this due to publish limitations on pypi: https://github.com/skodaconnect/myskoda/releases/tag/v0.17.1a1

fursov:
@WebSpider I tested this change/release overnight and this morning, different events arrived (including `charging-completed`), no errors seen. Seems this works and can be taken in use as is.

PR #312, opened 2025-01-04
Updated Enyaq 2025 to fixture from 0.16.2 lib

gschot:
Updated the fixture to the latest lib version as it was reset to the oldest version (i.e. no external temp) after fixture cleanup


dvx76 (approved):
Thanks for the update!

WebSpider:
Oooh, outside temperature for enyaq?!

gschot:
> Oooh, outside temperature for enyaq?!

It was already in there for a while, actually already in the fixture you deleted as duplicate but that was my bad for the unclear naming of the files I uploaded.

PR #321, opened 2025-01-07
add set air-conditioning timers feature

prvakt:
already tested by @WebSpider and working as expected

### DIFF
--- a/myskoda/myskoda.py
+++ b/myskoda/myskoda.py
@@ -30,9 +30,9 @@ from .event import Event
 from .models.air_conditioning import (
     AirConditioning,
     AirConditioningAtUnlock,
+    AirConditioningTimer,
     AirConditioningWithoutExternalPower,
     SeatHeating,
-    Timer,
     WindowHeating,
 )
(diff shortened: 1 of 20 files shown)


myskoda/myskoda.py
```diff
@@ -32,6 +32,7 @@
     AirConditioningAtUnlock,
     AirConditioningWithoutExternalPower,
     SeatHeating,
+    Timer,
```
WebSpider, line 35: I suspect multiple timers:

- AC Timer
- Charge Timer
- Possibly the aux heater has timer as well

Should we make this name more specific?

WebSpider (approved):
Cool addition! 👍

Please check if the list of used API endpoints needs updating, and a minor remark in the code about the name of the `Timer`.

Since it's nothing major, we can do both in follow-up PRs

myskoda/myskoda.py
prvakt: lets change it now as its not used anywhere yet

WebSpider (approved):
Lgtm

PR #323, opened 2025-01-10
make steeringWheelPosition and windowHeatingState optional

prvakt:
related to https://github.com/skodaconnect/homeassistant-myskoda/issues/548

### DIFF
--- a/myskoda/models/air_conditioning.py
+++ b/myskoda/models/air_conditioning.py
@@ -151,9 +151,11 @@ class AirConditioning(DataClassORJSONMixin):
     timers: list[AirConditioningTimer]
     errors: list[Any]
     state: AirConditioningState
-    steering_wheel_position: Side = field(metadata=field_options(alias="steeringWheelPosition"))
-    window_heating_state: WindowHeatingState = field(
-        metadata=field_options(alias="windowHeatingState")
+    steering_wheel_position: Side | None = field(
+        default=None, metadata=field_options(alias="steeringWheelPosition")
+    )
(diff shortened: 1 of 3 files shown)


WebSpider (commented):
This will do what's needed, but do we want to catch and model all error-states from API?
with "state" being "INVALID", this is clearly an error state.

prvakt:
> This will do what's needed, but do we want to catch and model all error-states from API? with "state" being "INVALID", this is clearly an error state.
> 
> Maybe it is a better idea to let the library throw a proper `MySkoda.SerializationError` exception in cases like this, so we can generally handle this in the integration and tell the user "whelp, we got a `SerializationError` and the state is `INVALID`, so something is wrong with MySkoda servers 🤷 "

Rest of the data from that class is already using Default = None so why not these 2? 

I agree that [#548](https://github.com/skodaconnect/homeassistant-myskoda/issues/548) was probably caused by problem on MySkoda side, but response still contains some usable data (timers have been returned) so why to throw it away?

WebSpider:
> Rest of the data from that class is already using Default = None so why not these 2? 
> 
> I agree that [#548](https://github.com/skodaconnect/homeassistant-myskoda/issues/548) was probably caused by problem on MySkoda side, but response still contains some usable data (timers have been returned) so why to throw it away?

Well, i guess the question is: If the API response indicates it's status is invalid, do we still trust the content of it? 

Let's go with this for now, but it may need to.change in the future. 

I agree we shouldnt change behaviour completely based on one occurance.

WebSpider (approved):
Lgtm

PR #348, opened 2025-02-04
Add support for mqtt departure-error-plug message

Giermann:
I discovered another error that was published via MQTT:
```
2025-01-30 17:30:34 python myskoda.mqtt[2702] DEBUG Message (service-event) received for TMBAXXXXXXXXXXXXX on topic departure: b'{"version":1,"traceId":"7b586354-fc7b-474a-bd6a-d69bbec13fa1","timestamp":"2025-01-30T16:30:30Z","producer":"SKODA_MHUB","name":"departure-error-plug","data":{"errorCode":"CLIMA","userId":"00000000-0000-0000-0000-000000000000","vin":"TMBAXXXXXXXXXXXXX"}}'
```
To add handling for this, I'd suggest to move the "charging-error" out of "models/charging.py" back into "models/service-event.py" and revert the adding of "error-code" to "ServiceEventChargingData" and to use a separate "ServiceEventErrorData" instead.

I'm not totally sure, why or when this error was fired - but maybe this opens the door to handle more MQTT published errors in future.

### DIFF
--- a/myskoda/models/charging.py
+++ b/myskoda/models/charging.py
@@ -60,10 +60,6 @@ class PlugUnlockMode(StrEnum):
     OFF = "OFF"
 
 
-class ServiceEventChargingError(StrEnum):
-    STOPPED_DEVICE = "STOPPED_DEVICE"
-
-
 @dataclass
 class Settings(DataClassORJSONMixin):
     available_charge_modes: list[ChargeMode] = field(
--- a/myskoda/models/service_event.py
+++ b/myskoda/models/service_event.py
@@ -8,7 +8,7 @@ from typing import Generic, TypeVar
 from mashumaro import field_options
 from mashumaro.mixins.orjson import DataClassORJSONMixin
 
-from .charging import ChargeMode, ChargingState, ServiceEventChargingError
+from .charging import ChargeMode, ChargingState
 
 
 class ServiceEventName(StrEnum):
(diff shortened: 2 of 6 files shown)


WebSpider:
Thanks for sharing your discovery!

I agree it's better to have all the service-event related code in one place, so moving the classes is fine.

I would like to take a minute to think about how we want to handle subclassing in the service-event structure, i feel we shouldnt let this grow only organically, but take a minute to think about the structure of the service events.

WebSpider (changes requested):
I agree with the proposed structure of the events.
Please fix the linter issues so we can move this forward.

WebSpider (changes requested):
Linter still fails on an import-soort and a test.

Tip: Run ruff locally so you can troubleshoot locally

Giermann:
@WebSpider I'd like to do all tests locally, but I have no clue how to run lint manually.
I even searched for the ability to run it at GitHub before requesting the review again, but I did not find anything.
And after committing the one change, the (failed) lint test was gone and not visible anymore.
I hope everything's fine now...

WebSpider:
No worries.

Tests and lint workflows are defined here: https://github.com/skodaconnect/myskoda/blob/main/.github%2Fworkflows%2Ftest.yaml

WebSpider (approved):
lgtm. 
Thanks @Giermann for your contribution :clinking_glasses:

PR #360, opened 2025-02-14
new fixture for Enyaq 85 MY2024

ExploracuriousAlex:


WebSpider (approved):
We already have this car model, but not at this software version (4.1).

Thanks for your contribution! :clinking_glasses:

PR #367, opened 2025-02-24
Support excluded_capabilities in get_vehicle

dvx76:
This will be used by the client to selectively exclude specific endpoints/capabilities, e.g. based on a timer.

This will replace the selective inclusion of the VEHICLE_HEALTH_INSPECTION capability in get_partial_vehicle() so this is removed here.

Required to resolve https://github.com/skodaconnect/homeassistant-myskoda/issues/468

### DIFF
--- a/myskoda/myskoda.py
+++ b/myskoda/myskoda.py
@@ -383,9 +383,11 @@ class MySkoda:
             return []
         return [vehicle.vin for vehicle in garage.vehicles]
 
-    async def get_vehicle(self, vin: str) -> Vehicle:
+    async def get_vehicle(
+        self, vin: str, excluded_capabilities: list[CapabilityId] | None = None
+    ) -> Vehicle:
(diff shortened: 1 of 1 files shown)


myskoda/myskoda.py
```diff
@@ -383,9 +383,11 @@ async def list_vehicle_vins(self) -> list[str]:
             return []
         return [vehicle.vin for vehicle in garage.vehicles]
 
-    async def get_vehicle(self, vin: str) -> Vehicle:
+    async def get_vehicle(
+        self, vin: str, excluded_capabilities: list[CapabilityId] | None
+    ) -> Vehicle:
```
WebSpider, line 388: Please add a default for backward compatibility

WebSpider (changes requested):
Looking good, just one minor remark: The new parameter doesn't have a default, which is causing a linter fail

WebSpider (approved):
lgtm

PR #378, opened 2025-03-06
Add example info about "Get ordered vehicle details" endpoint

lglerup:
Used to track order of a new vehicle.
Probably needs a few more examples later on, when orderStatus changes.
Assuming that would include some more information, if that has any interest as well 😄

### DIFF
--- a/docs/rest.md
+++ b/docs/rest.md
@@ -179,3 +179,73 @@
     ]
 }
 ```
+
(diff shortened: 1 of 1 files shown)


WebSpider (approved):
Thanks for documenting this! And yes, any new observations are always welcome, since devs do not have a car in all stages available :+1:

PR #379, opened 2025-03-06
Force new Vehicle to be returned

dvx76:
Ensure a new Vehicle object is returned by
get_vehicle/get_partial_vehicle.

If the same object is always returned, even if the content changes, the HA Coordinator doesn't know there was an update and thus doesn't trigger the entity listeners.

Fixes https://github.com/skodaconnect/homeassistant-myskoda/issues/651


dvx76:
> Lgtm. Did you get around to testing this?

Yes, I made the same change on my broken prod HA and it did the trick.

PR #394, opened 2025-03-23
Switch from poetry to uv

dvx76:
No real reason other than _I really wanted to try uv on a real project_.

Works well, but need to see if the actions and readthedocs still work. I don't use `just` myself, would you be able to check it @WebSpider ?

Usage is very similar to poetry:
* `poetry lock` -> `uv lock`
* `poetry install --all-extras` -> `uv sync --all-extras`
* `poetry run` -> `uv run`

```
❯ uv run myskoda  --user user@gmail.com --password '_passwd_' info TMBXXXXXXXXXXXX
capabilities:
  capabilities:
  - id: AUTOMATION
  ...
```

Dynamic version works as well

```
❯ uv build
Building source distribution...
Building wheel from source distribution...
Successfully built dist/myskoda-0.22.0.post1.dev0+66c8c91.tar.gz
Successfully built dist/myskoda-0.22.0.post1.dev0+66c8c91-py3-none-any.whl
```


WebSpider:
I dont have any special ties to poetry nor uv, and Homeassistant has chosen uv, so i'm not oppossed to trying it.

Interested to see how pypi publishing works, will look into it today.

pyproject.toml
```diff
@@ -1,84 +1,85 @@
...
 authors = [
-    "Frederick Gnodtke <frederick@gnodtke.net>",
-    "WebSpider <bacardicoke@gmail.com>",
-    "Fabrice Devaux <fabrice.devaux@gmail.com>",
+  {name = "Frederick Gnodtke", email = "frederick@gnodtke.net"},
+  {name = "WebSpider", email = "bacardicoke@gmail.com"},
+  {name = "Fabrice Devaux", email = "fabrice.devaux@gmail.com"}
 ]
 readme = "README.md"
+requires-python = ">=3.13.1"
```
WebSpider, line 11: Is this python version bump needed when just switching deps-managers?

.github/workflows/publish.yaml
```diff
@@ -19,26 +19,18 @@ jobs:
...
-
-      - name: Install dependencies
-        run: poetry install --sync --no-interaction
-
-      - name: Package project
-        run: poetry build
 
-      - name: Publish package distributions to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+      - name: Build and publisch package
```
WebSpider, line 33: Small typo

Justfile
WebSpider, line 1: I'm more than happy to ditch just, if we can swich to something like [pre-commit](https://pre-commit.com/) to handle this

WebSpider (dismissed):
Apart from the publish part, that we can only do in CI/CD, all seems to work fine. Let's see if we can get a prerelease published to pypi using UV.

Some small remarks, nothing big.

pyproject.toml
dvx76: It's not. I went with 3.13 because that's the dependency in homeassistant/core (https://github.com/home-assistant/core/blob/2025.3.4/pyproject.toml#L24), although it's 3.13.0 there, so I'll change it to that. Can keep 3.12.0 for now if you prefer.

(uv can, optionally, also install and manage Python itself)

.github/workflows/publish.yaml
dvx76: Fixed in next push. Thanks.

Justfile
dvx76: Also `uv` has a pre-commit hook to check the lockfile, like poetry has.

https://github.com/astral-sh/uv-pre-commit

pyproject.toml
WebSpider: In the homeassistant-myskoda integration, we depend on 3.13.0, as part of homeassistant 2025.3.2.
I personally prefer the library to have looser requirements than homeassistant, unless the code ends up being incompatible.

dvx76:
> Pretty confused why the CI job started failing 😕
> 
> ```
>       setuptools.errors.InvalidConfigError: Invalid dash-separated key
>       'python-tag' in 'bdist_wheel' (setup.cfg), please use the underscore
>       name 'python_tag' instead.
>       hint: This usually indicates a problem with the package or the build
>       environment.
>   help: `websockets` (v10.4) was included because `myskoda:dev` depends on
>         `amqtt` (v0.11.0b1) which depends on `websockets`
> ```

- https://github.com/pypa/setuptools/pull/4870#issuecomment-2748731521
- https://github.com/astral-sh/uv/issues/12440

pyproject.toml
dvx76: Makes sense. I've reverted to 3.12.0 here.

dvx76:
@WebSpider this should be stable now. After merging I'll do a pre-release to check the PyPI publish and readthedocs action.

WebSpider (approved):
Let's do this! 🚀

PR #404, opened 2025-03-25
Replace Just with pre-commit and direct commands

### LINKED ISSUE #400: Replace Just with pre-commit tasks
See  https://github.com/skodaconnect/myskoda/pull/394#discussion_r2009918630_

dvx76:
Rationale:
- One less thing to maintain
- Pre-commit does ruff and pyright. Locally before comitting + in CI
- Calling `pre-commit run` directly from CI action for now (over integrating with https://pre-commit.ci/)
- Unit tests are slower and undesirable to include in pre-commit. Can easily be run with `uv run pytest` (see also https://medium.com/@fistralpro/pytest-pre-commit-hook-b492edd0560e)

Fixes #400


WebSpider (approved):
Lgtm

WebSpider:
Do we maybe want to run pytest locally as well when we have changes in the tests/fixtures?

dvx76:
> Do we maybe want to run pytest locally as well when we have changes in the tests/fixtures?

I thought about this but ultimately a small change in any module can end up breaking tests so it's virtually impossible to define criteria to selectively run some or all tests _sometimes_. I also felt adding too many (slow-ish) pre-commit steps causes too much friction during development. I stumbled upon [this post](https://medium.com/@fistralpro/pytest-pre-commit-hook-b492edd0560e) which pushed me over the edge.

I do want to add a `CONTRIBUTING.md` document with some info on how to use `uv`, `pre-commit`, run tests, and what we generally expect from a PR. Will create an issue.

Also, I think pre-commit can trigger specific tasks on specific git hooks only, so it might be possible to auto-run unit tests but only on git push, not on every git commit. That would be reasonable I think.

PR #408, opened 2025-03-29
Fix type hint for mqtt callbacks

dvx76:
Just a minor type hinting fix I ran into while trying to understand something else, and learning more about async types.

`subscribe()` is passed an async function, which is callable returning a `Coroutine`, not an `Awaitable`, and definitely not `None`.

This then avoids the rogue cast to `Any` when passing the result of calling the callback function into `asyncio.create_task`.

### DIFF
--- a/myskoda/cli/mqtt.py
+++ b/myskoda/cli/mqtt.py
@@ -39,7 +39,7 @@ async def subscribe(ctx: Context) -> None:
     """Connect to the MQTT broker and listen for messages."""
     myskoda: MySkoda = ctx.obj["myskoda"]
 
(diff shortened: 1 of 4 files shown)


WebSpider (approved):
Lgtm.

Wondering if this means major work on the integration to type/call correctly there as well

dvx76:
> Lgtm.
> 
> Wondering if this means major work on the integration to type/call correctly there as well

Should not be the cast. The integration only calls `MySkoda.subscribe` once, and passes in an async function (`_on_mqtt_event`), which is what this code expects.

Issue #409, opened 2025-04-01
Finish MQTT event handling rework


dvx76:
In https://github.com/skodaconnect/myskoda/pull/407 and https://github.com/skodaconnect/homeassistant-myskoda/pull/737 we started reworking how MQTT events are handled, but only did it for charging events:

- `MySkoda` now owns the callback subscriptions and 'intercepts' the events so it can perform processing _before_ proxying the events to the subscribed callbacks
- Refresh API responses as required + update cached Vehicle with data from Event. All within `MySkoda` itself
- Client (coordinator in integration) upon receiving a charging event just needs to refresh the Vehicle using `MySkoda.vehicle()`

The same approach can be implemented for all other events.

Most likely requires implementing a debounce mechanism within `MySkoda` for querying the API endpoints since events all access events can happen quickly. Actually having a debounce on handling the event itself might be even better. This could also deduplicate the events being proxied back to the MySkoda client...

Also from https://github.com/skodaconnect/homeassistant-myskoda/pull/737#discussion_r2020964785

> We design a place for the client to know about what actually happened inside MySkoda while it was doing this unannounced update. Did we hit an API limit? Did the Internet connection break? Or was all fine. I don't think this is in this PR-set, but could be a follow-up.

WebSpider:
Moving event handling out of `myskoda.py` makes sense at this point.

dkadioglu:
Hi, I was just wondering what the state of this is? May I ask, when you plan to finish this?
Regardless of this, thank you for all your time and great work!

dkadioglu:
Thank you very much for your reply! By "no functional changes" you mean that, if I use the myskoda library in a project and when you finish this in the future, there should be no influence on 3rd party implementations using the myskoda library? I am just at the beginning of thinking about some ideas and would like to get a better understanding of possible implications.

WebSpider:
> Thank you very much for your reply! By "no functional changes" you mean that, if I use the myskoda library in a project and when you finish this in the future, there should be no influence on 3rd party implementations using the myskoda library? I am just at the beginning of thinking about some ideas and would like to get a better understanding of possible implications.

We do not expect any changes to the external API as a result of this rework.

PR #411, opened 2025-04-03
Debouncer + Refresh API data automatically on MQTT events

dvx76:
For #409

Putting this up as a draft to facilitate discussion.

This further moves API refreshes on operation and service MQTT events
from the integration (coordinator) to the MySkoda class.

I've implemented a simple debounce decorator. I'm using it slightly differently than what we do in the coordinator, for a reason. The debouncer is used on new methods which refresh a specific vehicle data (=~ endpoint). I want that to happen immediately because I don't want to block proxying the MQTT event to the client. I also want to maintain the 30s debounce/cooldown period between requests (to the same endpoint). Now, due to another event, the same refresh method is called again during those 30s, I still don't want to block for the same reason... So the only solution I could come up with is to drop those calls. That's what the `queue=False` parameter in the decorator does. Queuing up the call to execute at the end of the debounce period is pointless because by then we've already proxied the MQTT event so the coordinator won't know about the updated data.

In the coordinator this isn't a problem because the method we're debouncing comes _after_ receiving the event, and does both the refreshing (e.g. calls `myskoda.get_charging()`) _and_ the data update (`async_set_updated_data()`).

I think the solution might be to not proxy MQTT events after all and go back to directly having the `myskoda.MySkodaMqttClient` call the client callbacks. Then, the `MySkoda` class would offer a separate callback subscription for a new type of 'update' event. We can then have MySkoda independently refresh endpoint data and then notify the client, which, in the case of the coordinator, can then simply call `async_set_updated_data()`. The update event can optionally include details about what exactly was updated. I think this is what @WebSpider had in mind [here](https://github.com/skodaconnect/homeassistant-myskoda/pull/737#discussion_r2020964785).

I also have some reservations around refreshing the whole vehicle (all endpoints) on certain events (e.g. on error operations and access service events). See comments in code.

Side-note: I feel the `MySkoda` class is getting very big and bloated, making it hard to understand and work on. I plan to spend some time maybe this month to see about refactoring and simplifying it.


myskoda/myskoda.py
```diff
@@ -154,11 +155,66 @@ async def _on_mqtt_event(self, event: Event) -> None:
...
 
         self._emit_mqtt_event(event)
 
+    async def _process_operation_event(self, event: EventOperation) -> None:
+        """Refresh the appropriate vehicle data based on the operation details."""
+        _LOGGER.debug("Processing operation event: %s", event)
+        if event.operation.status == OperationStatus.IN_PROGRESS:
+            pass
+        elif event.operation.status == OperationStatus.ERROR:
+            pass  # Full update? Maybe instead just selective refresh as for completed operations?
```
WebSpider, line 184: We could update the parts that concern the just the `event.operation.operation` that is in error, yes

WebSpider (commented):
I think this is some solid improvements here. :tada: 

A few remarks:

- I agree that both the coordinator doing refresh calls and refreshing inside `MySkoda` is suboptimal. It will lead to multiple calls to the same endpoint where the chance of any significant change is minimal, but we do burn API credits.
- Adding this debouncer in `MySkoda` for MQTT is great, but why not expand it's use?

If the coordinator for some reason needs wants to do a refresh, why not let the coordinator re-use the debouncer that we added for MQTT? 

If the reason for the refresh is an MQTT-event that got passed up, that debouncer is probably either is currently in execution or awaiting it's cooldown because the data just got refreshed.

If the reason for the refresh is a periodic update, we're probably fine skipping the endpoints that are in cooldown, since the data there is fresh(-ish)

dvx76:
Thanks for looking at this and your feedback! I'm still chewing on this ... a lot. It's actually pretty complex to untangle when you think it trough and consider the different scenarios and corner cases.

I agree with this approach we'd want to get rid of the debouncer in the integration and only rely on the logic in MySkoda.

The approach I'm currently played around with is with a separate callback at the MySkoda level which notifies when any of the vehicle data (in MySkoda) has been updated. The coordinator would no longer directly request data (`get_charging_data()`, `get_vehicle()`, ...), except during startup. Instead it requests data to be updated (`refresh_charging_data()`, `refresh_vehicle()`, ... Those new methods are always debounced, which is now fine because the caller doesn't expect an immediate return value. At the end of those methods they notify the callback function of the coordinator, which just need to update it's data with what's in `self.myskoda.vehicles[vin]` at that moment.

The coordinator can still subscribe to MQTT events and update the last_event and last_operation sensors, but it no longer needs to do anything with those events. MySkoda handles processing the events and updating vehicle data accordingly, and when it's done it notifies the coordinator with the same callback as above.

dvx76:
Ok, next iteration is up. This goes with https://github.com/skodaconnect/homeassistant-myskoda/pull/746

This moves all update logic out of the coordinator and into the MySkoda lib.

The coordinator can still subscribe to MQTT events but they are not used for anything and just get exposed in corresponding entities.

The coordinator can now subscribe to 'update' events which MySkoda will callback when any Vehicle data was updated for whatever reason (including when it, the MySkoda lib, got an MQTT event). When receiving an update callback the coordinator just has to call async_set_updated_data and pass in the current Vehicle data.

WebSpider:
First of all: Thanks for your work! 🖖

I have yet to review the PR in detail, and plan to do so today. First thought is about backward compatibility: Do we change the behaviour of any of the methods/the vehicle state?

dvx76:
> First of all: Thanks for your work! 🖖
> 
> I have yet to review the PR in detail, and plan to do so today. First thought is about backward compatibility: Do we change the behaviour of any of the methods/the vehicle state?

Hah, I was _just_ doing some more re-organizing (just re-ordering methods, putting all public methods first, etc) and thinking about this. Yes, I think this should be considered a breaking change, although other users (as in, not our homeassistant-myskoda integration) wouldn't notice anything as all the typical `get_` methods still work the same. I think the only breaking change right now for other users would be the `subscribe` method is renamed to `subscribe_events`, but it still does the exact same thing. We now also have a `subscribe_updates` method which is handled directly by `MySkoda`.

I've prefixed the relevant commit with `BREAKING CHANGE:` (as per https://www.conventionalcommits.org) and am thinking maybe this is a good opportunity to finally promote the version to 1.0.0 😁

myskoda/myskoda.py
```diff
@@ -154,11 +155,66 @@ async def _on_mqtt_event(self, event: Event) -> None:
...
-        elif event.type == EventType.SERVICE_EVENT and event.topic == ServiceEventTopic.CHARGING:
-            await self._process_charging_event(event)
+        elif event.type == EventType.OPERATION:
+            await self._process_operation_event(event)
+        elif event.type == EventType.SERVICE_EVENT:
+            if event.topic == ServiceEventTopic.CHARGING:
+                await self._process_charging_event(event)
+            elif event.topic == ServiceEventTopic.ACCESS:
+                # NOTE: In the coordinator we are refreshing the full vehicle here...
+                # Do we really need to refresh the whole vehicle?
```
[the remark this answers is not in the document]
dvx76: I've reviewed my own logs a bit and I think on the one hand we get `access` event for locking/unlocking the car locally, which is totally expected. On the other hand I also get `access` events at the same time as other events, for example when starting/stopping AC, I get them together with `vehicle-awake` events.

I think refreshing the whole vehicle on access events is overkill. Just refreshing `Status` is the only thing directly relevant.

myskoda/myskoda.py
dvx76: Agreed. I'm changing this to still log a warning when we get this but then use the existing logic for other statuses to determine what to refresh.

myskoda/utils.py
```diff
@@ -0,0 +1,59 @@
...
+import functools
+from collections.abc import Awaitable, Callable
+from typing import ParamSpec
+
+# TODO @dvx76: Switch to Python 3.13 generics syntax once we're on 3.13
+# https://docs.python.org/3/library/typing.html#typing.ParamSpec
+
+P = ParamSpec("P")  # Represents the function parameters
+
+DEFAULT_DEBOUNCE_WAIT_SECONDS = 10.0
```
WebSpider, line 13: Shouldn't this be in `const.py`?

myskoda/myskoda.py
```diff
@@ -435,144 +474,121 @@ async def set_departure_timer(self, vin: Vin, timer: DepartureTimer) -> None:
...
         await future
 
-    async def get_departure_timers(self, vin: Vin, anonymize: bool = False) -> DepartureInfo:
-        """Retrieve departure timers for the specified vehicle."""
-        return (await self.rest_api.get_departure_timers(vin, anonymize=anonymize)).result
-
-    async def get_auth_token(self) -> str:
-        """Retrieve the main access token for the IDK session."""
-        return await self.rest_api.authorization.get_access_token()
+    async def refresh_user(self) -> None:
```
WebSpider, line 477: Is there a reason why this isn't debounced?

myskoda/myskoda.py
```diff
@@ -435,144 +474,121 @@ async def set_departure_timer(self, vin: Vin, timer: DepartureTimer) -> None:
...
-                    self._vehicles[vin].health = await self.get_health(vin)
-                case CapabilityId.DEPARTURE_TIMERS:
-                    self._vehicles[vin].departure_info = await self.get_departure_timers(vin)
-        except Exception as err:  # noqa: BLE001
-            _LOGGER.warning("Requesting %s failed: %s, continue", capa, err)
+        Get health only when missing and every 24h.
+        This avoids triggering battery protection, such as in Citigoe and Karoq.
+        https://github.com/skodaconnect/homeassistant-myskoda/issues/468
+        """
+        vehicle = self._vehicles[vin]
```
WebSpider, line 498: This could raise an exception when we refresh a vin that we don't know anything about yet

myskoda/myskoda.py
```diff
@@ -656,3 +672,183 @@ async def generate_get_fixture(
...
+        if event_data.charged_range:
+            charging_status.battery.remaining_cruising_range_in_meters = (
+                event_data.charged_range * 1000
+            )
+        if event_data.soc:
+            charging_status.battery.state_of_charge_in_percent = event_data.soc
+        if event_data.time_to_finish:
+            charging_status.remaining_time_to_fully_charged_in_minutes = event_data.time_to_finish
+        if event_data.state:
+            charging_status.state = event_data.state
```
WebSpider, line 830: This doesn't update any of the actual cached vehicle data, since this is a staticmethod that returns `None`

WebSpider (commented):
We're getting there!

myskoda/myskoda.py
dvx76: I can test it but I think it will, because it's being passed the `ChargingStatus` object of the current `self._vehicles[event.vin].charging`. I.e. it mutates the object being passed in. We could return it and have the caller store the return value back in `self._vehicles[event.vin].charging.status`, but it's effectively the same object.

myskoda/myskoda.py
dvx76: Yeah fair enough. From the coordinator it shouldn't happen because on startup we do a normal `get_partial_vehicle()` first. But I agree it should be made safe.

Changing this to call `get_vehicle()` if the VIN isn't known yet.

myskoda/myskoda.py
dvx76: Yes, it already has a custom cache logic built in so I didn't think there would be any point in debouncing it. If it gets called multiple times nothing will happen other than logging the debug message.

I can simplify the logic a bit though now that this is a separate function (it used to be baked into the coordinator)

On a sidenote, I wonder if I should have all the `get_` methods update the local vehicle object directly, rather than delegating that to the `refresh_` methods. If a client uses a `get_` method directly we'd still want the local vehicle object in MySkoda to be updated I think.

E..g. 

```python
    async def get_charging(self, vin: Vin, anonymize: bool = False) -> Charging:
        """Retrieve information related to charging for the specified vehicle."""
        result = (await self.rest_api.get_charging(vin, anonymize=anonymize)).result
        if vehicle := self._vehicles.get(vin):
            vehicle.charging = result
        return result

    @async_debounce(immediate=True)
    async def refresh_charging(self, vin: Vin, notify: bool = True) -> None:
        """Refresh charging data for the provided Vin."""
        await self.get_charging(vin)
        if notify:
            self._notify_callbacks(vin)
```

myskoda/utils.py
dvx76: Makes sense. Moved.

WebSpider (dismissed):
Looks solid! :muscle: 

Maybe we need to update the readme a bit (https://github.com/skodaconnect/myskoda/tree/main/docs#basic-example) and readthedocs may need some love as well, otherwise looking good!

dvx76:
> Looks solid! 💪
> 
> Maybe we need to update the readme a bit ([`main`/docs#basic-example](https://github.com/skodaconnect/myskoda/tree/main/docs#basic-example)) and readthedocs may need some love as well, otherwise looking good!

Awesome. I've fixed up the basic-example section - it's actually copy-pasteable now. The additional docs going into readthedocs were still mostly accurate because all the `get_*` methods haven't really changed. I just fixed the event subscription.

I'm going to merge this now before it gets stale and so I can get a passing pipeline on the corresponding homeassistant-myskoda PR.

I feel like this is maybe a good time to bump the version of this package to 1.0.0, what do you think?

WebSpider:
Yeah im fine with a 1.0, we should definately also update readthedocs then

dvx76:
> Yeah im fine with a 1.0, we should definately also update readthedocs then

Sounds good. I've made some additional updates, and added a new file specifically around refreshing data and the new `subscribe_updates()` callback registration. I've also added the `subscribe()` method back in. It just calls `subscribe_events()` and logs a deprecation warning. We can then remove it later.

I'll do a little bit more validation this morning and then merge and release so I can get the PR for the integration in shape. 🚀

PR #413, opened 2025-04-14
Fix: user.nickname is optional

dvx76:
Fixes https://github.com/skodaconnect/homeassistant-myskoda/issues/747


WebSpider (approved):
Lgtm.

I think ive also seen that first_name can be empty in one of the logs that have passed, so i guess its entirely optional by now.

PR #419, opened 2025-04-25
Add new fixture for Enyaq 85, 2025 facelift

macfreek:
A fixture of my vehicle. Fresh from the press, err, factory (<100 km on the road). A bit similar to existing fixtures (the Škoda Enyaq 85 is already twice in the list of fixtures), so feel free to reject it.

I've submitted it anyway, because I noticed a few uncommon values:
* The existing Enyaq model ID 5AZJK2 has `model_year: 2024`, `software_version: 4.1` and `trim_level: 85`. I kind of expect my result to have `model_year: 2025` (as this is the 2025 facelift), but surprisingly it reports `model_year: 2026` (!), `software_version: 5.4` and `trim_level: 'Business Edition'`. The developer of a MySkoda client for the Homey stopped support a month ago, because "VW Group changes their API every month". With these inconsistent return values, I start to see why.
* There is no fixture yet with `state: "GUEST_USER"` (only `state: "ACTIVATED"`), and with that there is no fixture yet with result `INSUFFICIENT_RIGHTS` for `PLUG_AND_CHARGE` status.
* This fixture is the first with `next_charging_time: null` in the charging profile.

There are some other small tidbits, but nothing more that is not already present in another fixture.


WebSpider:
Thanks for your fixture!

Actually, we've seen model years appear before the start of the new year, but april is a but early, i admit :)

With the experience we had from the skodaconnect integration into HA, we chose a scalable setup, so that usually only minor changes are needed to adopt new cars/models/modelyears and this has served us well so far.

If the developer of the Homey plugin/addon would like to collaborate, they are welcome in this project. 👍

WebSpider (approved):
Thanks for your contribution!

PR #420, opened 2025-04-26
Perform full vehicle refresh on access events

dvx76:
This is annoying because access events seems to occur for about anything but at least position is something we need to refresh on access it seems. Departure events explicitly refresh position but those don't seem to occur reliably - or perhaps occur before the API knows the new position...

Note that a full refresh on access events is what we already had prior to moving event handling from the coordinator into the myskoda lib, so this is actually just reverting back to previous behavior.

Perhaps we'll find a better event to trigger a full refresh (or a position refresh).


WebSpider (approved):
This probably will do it until we find a better event.

PR #423, opened 2025-05-04
Add start/stop ventilation (air-conditioning)

BenjaminHae:
Implementing ventilation according to https://github.com/skodaconnect/myskoda/blob/main/docs/api_endpoints.md


BenjaminHae:
I have an Octavia from 2023.

For some reason the myskoda cli does not come back after the command. But I'm very confident that the MQTT message it's waiting for is correct.

During my debugging sessions, no messages other than the WAKE_UP did appear.

Btw: I'm doing this, because this project is the best reference implementation for the MySkoda API, actually my goal was to implement it for iobroker (which I've done [here](https://github.com/TA2k/ioBroker.vw-connect/pull/370)).

dvx76 (approved):

PR #429, opened 2025-05-09
fix: position.address is now optional

dvx76:
Fixes https://github.com/skodaconnect/homeassistant-myskoda/issues/787


WebSpider (approved):
Lgtm

PR #431, opened 2025-05-14
Add fixture for Elroq 85

stephanme:


WebSpider (approved):
Thanks a lot for this fixture, we dont have many Elroq on file yet! 🎉

PR #436, opened 2025-05-20
chore/refactor: remove amqtt, refactor tests + MySkodaMqttClient

### LINKED ISSUE #401: Replace amqtt
amqtt is used for unit tests but is old and unmaintained. It has an upper bound dependency on an old version of websockets, which breaks when build with a setuptools with this change: https://github.com/pypa/setuptools/pull/4870#issuecomment-2748731521

What we want: find a recent and maintained replacement for amqtt, or implement the related unit tests differently, so we get rid of these old brittle dependencies.

dvx76:
### Decouple `aiomqtt.Client `from `MySkodaMqttClient`

Allow tests injecting their own fake client so that we can test
`MySkodaMqttClient` without needing an actual MQTT broker.

### Remove `amqtt`, refactor affected tests

Fixes #401.

amqtt is old, unmaintained, and causes dependency issues.

Rather than trying to replace it we've refactored the tests and tested
code to no longer depend on a 'real' MQTT broker running. This focuses
the tests on our own code instead of mostly testing the MQTT client lib
(`aiomqtt`).

This is done by injecting a fake `aiomqtt.Client` in `MySkodaMqttClient`.

The tests in `tests/test_operations.py` can probably be further simplified/de-duplicated but that will be for another day and not what I was going for here 😅 .

### DIFF
--- a/myskoda/auth/authorization.py
+++ b/myskoda/auth/authorization.py
@@ -51,6 +51,10 @@ class IDKSession(DataClassORJSONMixin):
 class Authorization(ABC):
     """Class that holds Authorization information and authorization state of the session."""
 
+    # TODO @dvx75: I think this should be much more abstract and void of most concrete
+    # implementations. At least not have dependencies on ClientSession.
+    # Maybe also move to Protocol instead of ABC.
+
     session: ClientSession
     idk_session: IDKSession | None = None
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,25 +1,137 @@
 """Test helpers."""
 
+import asyncio
+import copy
 import json
-from asyncio.timeouts import timeout
 from collections.abc import AsyncIterator, Generator
+from contextlib import AbstractAsyncContextManager
 from pathlib import Path
 from socket import socket
+from types import TracebackType
+from typing import Any
 from unittest.mock import AsyncMock
 
+import aiomqtt
 import pytest
 from aiohttp import ClientSession
 from aioresponses import aioresponses
-from amqtt.broker import Broker, CancelledError
-from amqtt.client import MQTTClient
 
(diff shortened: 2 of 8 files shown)


tests/conftest.py
```diff
@@ -102,25 +169,23 @@ def mock_default_routes(responses: aioresponses) -> None:
 
 
 @pytest.fixture
-async def myskoda(responses: aioresponses, broker_port: int) -> AsyncIterator[MySkoda]:
+async def myskoda(
```
WebSpider, line 172: We probably should avoid naming things just `myskoda`. It causes confusion what we're doing, since just about anything in this package can be named `myskoda`.

WebSpider (approved):
Nice job! :muscle: 
Some inline comments, but nothing major.

myskoda/auth/authorization.py
```diff
@@ -51,6 +51,10 @@ class IDKSession(DataClassORJSONMixin):
 class Authorization(ABC):
     """Class that holds Authorization information and authorization state of the session."""
 
+    # TODO @dvx75: I think this should be much more abstract and void of most concrete
+    # implementations. At least not have dependencies on ClientSession.
+    # Maybe also move to Protocol instead of ABC.
+
```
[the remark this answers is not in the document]
dvx76: Yeah. I'd be interested to look at that since I haven't had much exposure to oidc so would be good to learn. Did you want to see an issue open for this?

tests/conftest.py
dvx76: Heh yeah I know what you mean. This is an existing fixture used in a whole bunch of tests so if we do change it I'd rather keep it out of this PR. Maybe `myskoda_test_instance`, although that's a little long maybe.

Issue #444, opened 2025-06-04
Improve unit tests for myskoda.py


dvx76:
We only have ~58% coverage in myskoda.py which is actually the external interface of the whole package.

dvx76:
Duplicate of #410 . Getting older is so fun.

PR #445, opened 2025-06-04
feat: refresh charging on charging events

dvx76:
If the event is a charging event and not a change-soc event (which already contains data itself) force a refresh of the charging endpoint data.

On Citigoe people seem to only receive `charging-status-changed` events, which contain no usable data. Newer EVs (Enyaq, ...) don't see that event at all and instead get `change-soc`.

Fixes: skodaconnect/homeassistant-myskoda#821

Note: working on this I noticed we're pretty light on unit tests against myskoda.py specifically and there was nothing existing I could easily update or extend to cover the new code here. So I opened #444 444


WebSpider (approved):
Lgtm

PR #448, opened 2025-06-11
chore!: remove unused ssl_context

dvx76:
BREAKING CHANGE: This changes the MySkoda init arguments. `ssl_context` should no longer be set.

This doesn't do anything anymore since asyncio-paho was replaced with aiomqtt in skodaconnect/myskoda#140

Depends on skodaconnect/homeassistant-myskoda/pull/833


WebSpider (commented):
If we allow overflow `**kwargs`, we dont need to make this breaking.

PR #449, opened 2025-06-12
fix: pass the HA ssl_context into the mqtt client

dvx76:
Used to be the case but was accidentally dropped when migrating to aiomqtt.


WebSpider (approved):
Lgtm

PR #451, opened 2025-06-12
fix: Booking.closed_date is optional

dvx76:
For `activeBookings` the `closedDate` can be missing.

```
  File "<string>", line 8, in __mashumaro_from_dict_json__
  File "<string>", line 21, in __mashumaro_from_dict_json__
mashumaro.exceptions.MissingField: Field "closed_date" of type datetime is missing in Booking instance
```


WebSpider (approved):
Lgtm

PR #485, opened 2025-10-04
test: update test_generated to force loading fixtures in utf-8 encoding

bassduh:
pytest was returning errors converting Škoda in the title on Windows OS.

```
  -         'title': 'Å\xa0koda Octavia',
  ?                   ^^^^^
  +         'title': 'Škoda Octavia',
```
  
Add some lines in the contribution document, because on Windows there are some differences for the setup of UV.


WebSpider (approved):
Personally i do everything on Linux, so awesome we now have Windows developers as well 💪

Thanks for you contribution 👍

PR #496, opened 2026-01-17
Added support for single-trips endpoint

moledzki:
Hi,

I've added support for single-trips endpoint.

### DIFF
--- a/myskoda/anonymize.py
+++ b/myskoda/anonymize.py
@@ -1,47 +1,25 @@
 """Methods for anonymizing data from the API."""
 
-import datetime
 import re
-
-from myskoda.models.air_conditioning import AirConditioning
-from myskoda.models.auxiliary_heating import AuxiliaryHeating
-from myskoda.models.charging import Charging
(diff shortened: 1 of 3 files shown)


myskoda/cli/requests.py
```diff
@@ -199,6 +199,42 @@ async def trip_statistics(ctx: Context, vin: str, anonymize: bool) -> None:
...
+    start: datetime | None = None,
+    end: datetime | None = None,
+) -> None:
+    """Retrieve detailed statistics about past trips.
+    If you want to filter by date, provide both start and end date."""
+    if (start is None) ^ (end is None):
+        err_msg = "Both --start and --end must be provided."
+        raise click.BadParameter(err_msg)
+
+    kwargs: dict[str, Any] = {
```
WebSpider: Please make this a model as well, maybe introduce a `SingleTrip`?

moledzki:
> I will run this branch this weekend and see if it works for my car.

Can you check get_status endpoint also? It seems that overall section has new field: reliableLockStatus

Issue #501, opened 2026-02-10
Provide VIN to callback(s) registered with subscribe_updates()


Giermann:
I just noticed the introduced cache for vehicles and the corresponding refresh_* routines and subscribe_updates.
But when I tried to use this in my application, I found a big disadvantage: the callback is being called without any reference to the updated VIN!
This might be sufficient for observing a single static VIN, but with a small change it would be much more usable.

I will provide a PR, but wanted to create the issue first.

PR #502, opened 2026-02-10
Provide VIN to callback(s) registered with subscribe_updates()

Giermann:
See issue #501 for details.

### DIFF
--- a/myskoda/myskoda.py
+++ b/myskoda/myskoda.py
@@ -872,15 +872,7 @@ class MySkoda:
     def _notify_callbacks(self, vin: Vin) -> None:
         """Execute registered callback functions for the vin."""
         for callback in self._callbacks.get(vin, []):
-            try:
-                result = callback(vin)
-            except TypeError:
-                """For backwards compatibility only."""
-                _LOGGER.warning(
-                    "Callbacks registered with subscribe_events() receive the VIN "
-                    "as parameter now, please update your code."
-                )
-                result = callback(vin)
+            result = callback(vin)
             if result is not None:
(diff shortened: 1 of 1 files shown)


WebSpider:
I had a look at this piece of the code, since it's been a while :)

Basically, the main difference between `subscribe_events` and `subscribe_updates` is that the latter only gives you a signal that the vehicle has been updated, but since we do not know what exactly has been updated, we leave that to the callback to handle.

`subscribe_events` passes the details of the event right along, since we have it and we know it's complete, there is no other source of info for this event.

Since we are trying to solve the fact that the `Vin` data is missing when processing a callback, would it solve your problem if we optionally accept `Vin` as an argument, and just pass it back if we receive it, instead of just the signal that something has changed?

Giermann:
Still, I am not sure wheter we should really simply provide the VIN to the callbacks.

WebSpider wrote (although he later revised this comment):
"Basically, [...] subscribe_updates [...] only gives you a signal that the vehicle has been updated, but since we do not know what exactly has been updated, we leave that to the callback to handle."

But if the callbacks are interested in the trigger of that change, we could exented _notify_callbacks() to receive and publish that information too. If this is required (in the future?), we should introduce a BaseEvent parameter now instead of a simple VIN, which could be extended later.
Otherwise there would be another change of callback functions...

But I am fine with both - as long as the callback receives *any* information about the affected VIN.

Giermann:
I overlooked that, only noticed the now still present "Expected 1 more positional argument (reportCallIssue)", which is by design.

PR #543, opened 2026-03-27
feat: add support of vehicle-information endpoint

moledzki:
Hi,
I've added support of /v1/vehicle-information/{vin} endpoint. Please review it.


myskoda/models/vehicle_info.py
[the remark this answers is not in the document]
moledzki: Thanks for pointing it out – I missed that.

I'm wondering why the `Gearbox` class isn't used in the `Specification` class. Was there any reason not to include a gearbox field there? 
Another question – can I change the `view_point` field in the `Render` class to `StrEnum`, the same way as `ViewType`?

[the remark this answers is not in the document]
moledzki: I'm asking about Specification class, because after adding optional gearbox field to Specification class a lot of test_generated tests needs to be fixed and it will be better not to fix all these tests for nothing :)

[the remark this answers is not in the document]
WebSpider: You are right to wonder, and I think this is an omission. Gearbox is part of the `Specification` api-response and this should be in there.

Yes, this will cause changes to the fixtures, and yes ... that is a bit of work. Been there, done that :)

myskoda/models/info.py
```diff
@@ -242,6 +242,7 @@ class Specification(DataClassORJSONMixin):
     model_year: str = field(metadata=field_options(alias="modelYear"))
     system_code: str = field(metadata=field_options(alias="systemCode"))
     system_model_id: str = field(metadata=field_options(alias="systemModelId"))
+    gearbox: Gearbox | None = field(default=None, metadata=field_options(alias="gearbox"))
```
WebSpider: Nit: when the alias and the field are identical, you can drop the alias.

myskoda/models/info.py
```diff
@@ -265,13 +266,32 @@ class ServicePartner(DataClassORJSONMixin):
...
+    UNMODIFIED_INTERIOR_FRONT = "UNMODIFIED_INTERIOR_FRONT"
+    UNMODIFIED_INTERIOR_BOOT = "UNMODIFIED_INTERIOR_BOOT"
     HOME = "HOME"
     CHARGING_LIGHT = "CHARGING_LIGHT"
     CHARGING_DARK = "CHARGING_DARK"
     PLUGGED_IN_DARK = "PLUGGED_IN_DARK"
     PLUGGED_IN_LIGHT = "PLUGGED_IN_LIGHT"
 
 
+class ViewPoint(StrEnum):
```
WebSpider: If you make it a `CaseInsensitiveEnum`, you dont need to duplicate

myskoda/models/vehicle_info.py
```diff
@@ -0,0 +1,49 @@
...
+
+from mashumaro import field_options
+from mashumaro.mixins.orjson import DataClassORJSONMixin
+
+from .common import BaseResponse
+from .info import CompositeRender, Render, Specification
+
+
+@dataclass
+class VehicleInfo(DataClassORJSONMixin):
```
WebSpider: I checked, and we already have all this information in the `Info` class, why duplicate this?

myskoda/models/vehicle_info.py
```diff
@@ -0,0 +1,49 @@
...
+    vehicle_specification: Specification = field(
+        metadata=field_options(alias="vehicleSpecification")
+    )
+    composite_renderers: list[CompositeRender] = field(
+        metadata=field_options(alias="compositeRenders")
+    )
+
+
+@dataclass
+class VehicleRenders(BaseResponse):
```
WebSpider, line 22: This information is also already present in the `Info` class.

myskoda/myskoda.py
```diff
@@ -402,6 +403,25 @@ async def get_health(self, vin: Vin, anonymize: bool = False) -> Health:
...
+    async def get_vehicle_equipment(self, vin: Vin, anonymize: bool = False) -> VehicleEquipment:
+        """Retrieve vehicle equipment information for the specified vehicle."""
+        return (await self.rest_api.get_vehicle_equipment(vin, anonymize=anonymize)).result
+
+    async def get_vehicle_full_info(self, vin: Vin, anonymize: bool = False) -> VehicleFullInfo:
+        """Retrieve all available information for the specified vehicle."""
+        vehicle_info = await self.get_vehicle_info(vin, anonymize=anonymize)
+        equipment = await self.get_vehicle_equipment(vin, anonymize=anonymize)
+        renders = await self.get_vehicle_renders(vin, anonymize=anonymize)
+        return VehicleFullInfo(info=vehicle_info, equipment=equipment, renders=renders)
```
WebSpider: This appears to be a convenience method for getting the base car info. I don't mind adding this, but could you explain why calling the current `MySkoda.get_info()` combined with your suggested `MySkoda.get_vehicle_equipment()` doesn't work for you?

WebSpider (commented):
I have reviewed this PR and where I see it really adds features we do not have yet (specifically, the equipment), it also seems to duplicate some existing features in a slightly different way.

Could you please explain why you propose to add them, as the current methods:

- Use only 1 API call
- Retrieve the same information

myskoda/models/info.py
moledzki: OK, I'll change this

myskoda/models/info.py
moledzki: `CaseInsensitiveEnum` doesn't solve all issues, `test_generated` are still failing. I'm trying to find the best way to fix this.

myskoda/models/vehicle_info.py
moledzki: I didn't noticed this, this class will be removed

myskoda/models/vehicle_info.py
moledzki: Yes, but this endpoint provides more renders, eg. INTERIOR_BOOT which are not visible in `get_info` response.

myskoda/myskoda.py
moledzki: `MySkoda.get_info()`, does not provide all renders.

moledzki:
> I have reviewed this PR and where I see it really adds features we do not have yet (specifically, the equipment), it also seems to duplicate some existing features in a slightly different way.
> 
> Could you please explain why you propose to add them, as the current methods:
> 
> * Use only 1 API call
> * Retrieve the same information

My goal was to implement the comprehensive `vehicle-information` endpoint, and I didn't notice that some of the data was already available via the `get_info` method. Additionally, I assumed that one of the project's goals was to wrap all available REST endpoints.

`get_vehicle_full_info` was added as a convenience method to gather data from the entire `vehicle-information` tree. If you prefer not to have methods that provide redundant information, I can remove `get_vehicle_full_info`.

However, I think the `get_vehicle_renders` method should be kept, as it provides additional renders. The open question is whether these new renders (and perhaps the equipment data) should be merged into the `Info` class.

WebSpider:
> > I have reviewed this PR and where I see it really adds features we do not have yet (specifically, the equipment), it also seems to duplicate some existing features in a slightly different way.
> > Could you please explain why you propose to add them, as the current methods:
> > 
> > * Use only 1 API call
> > * Retrieve the same information
> 
> My goal was to implement the comprehensive `vehicle-information` endpoint, and I didn't notice that some of the data was already available via the `get_info` method. Additionally, I assumed that one of the project's goals was to wrap all available REST endpoints.
> 
> `get_vehicle_full_info` was added as a convenience method to gather data from the entire `vehicle-information` tree. If you prefer not to have methods that provide redundant information, I can remove `get_vehicle_full_info`.
> 
> However, I think the `get_vehicle_renders` method should be kept, as it provides additional renders. The open question is whether these new renders (and perhaps the equipment data) should be merged into the `Info` class.

Yeah I understand, and you are right. It doesn't hurt to provide these new methods at all, as long as we can re-use code where this is appropriate. This way, we can supply as much information about the vehicle and interaction with the vehicle as possible. :heart: 

The convenience method is not a problem, you don't have to remove it!

myskoda/models/vehicle_info.py
moledzki: It seems that this class will be needed because the `Info` class has a few non-optional fields which are not returned by this endpoint. Unfortunately, making them optional will not solve all the issues, because in the `vehicle-info` endpoint, the `Specification` object is returned in the `vehicleSpecification` field instead of the `specification` field.

myskoda/models/info.py
moledzki: I don't know how to remove these duplicates. Adding `CaseInsensitiveEnum` will only fix parsing issues, but doesn't fix `test_generated` tests.
The only way to fix these tests is to add lowercase entries to this enum, or fix all input data for `test_generated` tests.
What should I do?

myskoda/rest_api.py
```diff
@@ -352,6 +353,48 @@ async def get_health(self, vin: str, anonymize: bool = False) -> GetEndpointResu
         url = anonymize_url(url) if anonymize else url
         return GetEndpointResult(url=url, raw=raw, result=result)
 
+    async def get_vehicle_info(
```
dvx76, line 359: nit: can we place these new methods together with the existing `get_software_update_status()` so all the `/v1/vehicle-information/` are together. Preferably in this order:

- `get_vehicle_info()`
- `get_software_update_status()`
- `get_vehicle_renders()`
- `get_vehicle_equipment()`

myskoda/rest_api.py
```diff
@@ -352,6 +353,48 @@ async def get_health(self, vin: str, anonymize: bool = False) -> GetEndpointResu
...
+        self, vin: str, anonymize: bool = False
+    ) -> GetEndpointResult[VehicleInfo]:
+        """Retrieve vehicle info for the specified vehicle."""
+        url = f"/v1/vehicle-information/{vin}"
+        raw = self.process_json(
+            data=await self._make_get_request(url),
+            anonymize=anonymize,
+            anonymization_fn=anonymize_health,
```
dvx76: Even though most of these `anonymize` functions do nothing, using `anonymize_health` here seems semantically incorrect.

Suggestion: rename the existing `anonymize_software_update_status()` to `anonymize_vehicle_information()` and use that for these 3 new methods + the existing `get_software_update_status()`

myskoda/myskoda.py
```diff
@@ -402,6 +403,25 @@ async def get_health(self, vin: Vin, anonymize: bool = False) -> Health:
         """Retrieve health information for the specified vehicle."""
         return (await self.rest_api.get_health(vin, anonymize=anonymize)).result
 
+    async def get_vehicle_info(self, vin: Vin, anonymize: bool = False) -> VehicleInfo:
```
dvx76, line 406: Similar comment, can we group these with the existing `get_software_update_status()`?

- `get_vehicle_info()`
- `get_software_update_status()`
- `get_vehicle_renders()`
- `get_vehicle_equipment()`

myskoda/models/vehicle_info.py
```diff
@@ -0,0 +1,49 @@
...
+@dataclass
+class VehicleEquipment(BaseResponse):
+    equipment: list[Equipment] = field(metadata=field_options(alias="equipment"))
+
+
+@dataclass
+class VehicleFullInfo(DataClassORJSONMixin):
+    info: VehicleInfo = field(metadata=field_options(alias="info"))
+    equipment: VehicleEquipment = field(metadata=field_options(alias="equipment"))
+    renders: VehicleRenders = field(metadata=field_options(alias="renders"))
```
dvx76: Wondering if this should include `SoftwareUpdateStatus`, since it's also sourced from a `/v1/vehicle-information/` endpoint 🤔

myskoda/models/vehicle_info.py
dvx76: `Info` and `VehicleInfo` can both subclass a common base class. Should prob start from `BaseResponse` as well. E.g.

```python
@dataclass
class InfoBase(BaseResponse):
    device_platform: str = field(metadata=field_options(alias="devicePlatform"))
    renders: list[Render]
    composite_renders: list[CompositeRender] = field(
        metadata=field_options(alias="compositeRenders")
    )

@dataclass
class VehicleInfo(InfoBase):
    vehicle_specification: Specification = field(
        metadata=field_options(alias="vehicleSpecification")
    )

@dataclass
class Info(InfoBase):
    state: VehicleState
    specification: Specification
    # ...
```

myskoda/models/vehicle_info.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@dataclass
+class VehicleInfo(DataClassORJSONMixin):
+    device_platform: str = field(metadata=field_options(alias="devicePlatform"))
+    renders: list[Render] = field(metadata=field_options(alias="renders"))
+    vehicle_specification: Specification = field(
+        metadata=field_options(alias="vehicleSpecification")
+    )
+    composite_renderers: list[CompositeRender] = field(
```
dvx76: typo


```suggestion
    composite_renders: list[CompositeRender] = field(
```

myskoda/models/vehicle_info.py
```diff
@@ -0,0 +1,49 @@
...
+        metadata=field_options(alias="compositeRenders")
+    )
+
+
+@dataclass
+class VehicleRenders(BaseResponse):
+    renders: list[Render] = field(metadata=field_options(alias="renders"))
+    composite_renderers: list[CompositeRender] = field(
```
dvx76: typo

```suggestion
    composite_renders: list[CompositeRender] = field(
```

tests/test_rest_api.py
```diff
@@ -586,3 +586,158 @@ def assert_driving_score_result(
...
+        assert gearbox is not None
+        assert gearbox.type == gearbox_json["type"]
+
+        composite_renderers = get_vehicle_info_result.composite_renderers
+        composite_renderers_json = vehicle_information_json["compositeRenders"]
+        assert len(composite_renderers) == len(composite_renderers_json)
+        for i in range(len(composite_renderers)):
+            assert composite_renderers[i].view_type == composite_renderers_json[i]["viewType"]
+            layers = composite_renderers[i].layers
+            layers_json = composite_renderers_json[i]["layers"]
```
dvx76: typos


```suggestion
        composite_renders = get_vehicle_info_result.composite_renderers
        composite_renders_json = vehicle_information_json["compositeRenders"]
        assert len(composite_renders) == len(composite_renders_json)
        for i in range(len(composite_renders)):
            assert composite_renders[i].view_type == composite_renders_json[i]["viewType"]
            layers = composite_renders[i].layers
            layers_json = composite_renders_json[i]["layers"]
```

myskoda/models/info.py
dvx76: How do the tests fail? `CaseInsensitiveStrEnum` should work IMO.

But from the fixtures it looks like the actual values are always lowercase? So why not only have lowercase values here?

```python
class ViewPoint(StrEnum):
    EXTERIOR_SIDE = "exterior_side"
    # ...
```

dvx76 (changes requested):
Thanks for helping with this project. With a few tweaks this should be a good addition.

myskoda/rest_api.py
moledzki: I forgot to change this after copy-paste operation. I will fix this

myskoda/models/vehicle_info.py
moledzki: OK, I will update this.

myskoda/rest_api.py
moledzki: OK

myskoda/myskoda.py
moledzki: OK

myskoda/models/vehicle_info.py
moledzki: OK, I will fix this

tests/test_rest_api.py
moledzki: OK, I will fix this.

myskoda/models/info.py
moledzki: Tests are failing because during parsing lowercase string we are returning uppercase version of enum. And then we are comparing lowercase string to string representation of uppercase enum.

myskoda/models/vehicle_info.py
moledzki: It should work - I will fix this.

myskoda/models/info.py
moledzki: Using lowercase values also doesn't solve all problems, two tests are still failing.

moledzki:
I've submitted version with "duplicated" entries in ViewPoint enum. I can't fix this issue without changing test inputs, and I'm not sure if we want to do this.

myskoda/models/info.py
dvx76: Ok, so the actual problem is that the old `api/v2/garage/vehicles/{vin}` returns `viewType` values lowercase, whereas this new `api/v1/vehicle-info` returns them uppercase.

So either we treat them as separate entities, or we standardize on one of the two for our internal representation (models). Since these seem to really represent the same thing, I'd rather standardize.

With the `viewType` being uppercase for both endpoints, let's standardize on uppercase.

This means the previously generated fixtures for the garage endpoint
(which treated `viewPoint` as a base str) need to be updated.

I'll push a commit with what I mean.

dvx76 (approved):
LGTM in this form, @WebSpider feel free to merge if you are good with it.

WebSpider (commented):
I am *almost* happy, thanks for putting this time and effort in! 

Due to the changes in the `Info` class, this PR is currently breaking, since anyone expecting the `Info` class old structure will have to update their implementation. 

IMO, this is not needed. If we keep the `Info` class as-is, but mark it deprecated and scheduled for removal in 3.0.0, we can go ahead now and merge this. 

For now, this can be done by a simple comment in the class, possibly with an additional logger as we get closer to 3.0.0

dvx76:
> Due to the changes in the `Info` class, this PR is currently breaking, since anyone expecting the `Info` class old structure will have to update their implementation.

To be clear, unless I misunderstand, the only change in `Info` that clients would notice is that `renders` is now a list of `Render` `StrEnum`'s, instead of plain `str`'s. That _could_ still break something for someone but I don't consider it a big change. In most cases a `StrEnum` and a `str` can be used interchangeably.

Why would we deprecate `Info` ?

WebSpider:
> > Due to the changes in the `Info` class, this PR is currently breaking, since anyone expecting the `Info` class old structure will have to update their implementation.
> 
> To be clear, unless I misunderstand, the only change in `Info` that clients would notice is that `renders` is now a list of `Render` `StrEnum`'s, instead of plain `str`'s. That _could_ still break something for someone but I don't consider it a big change. In most cases a `StrEnum` and a `str` can be used interchangeably.
> 
> Why would we deprecate `Info` ?
> 

The mobile view of github messed me up here, i missed the subclassing. You are right, this change in Renders isnt quite as breaking as what i saw. I'll merge and make a note in the release notes.

PR #550, opened 2026-03-30
Support for setting preferred charging times

max-2468:
Hi all,

as I really would like to charge my Enyaq in an underground garage cheaply at some point without having to walk there every day I've added support to the myskoda library to set the preferred charging times. This is a start address [#685](https://github.com/skodaconnect/homeassistant-myskoda/issues/685) and [#866](https://github.com/skodaconnect/homeassistant-myskoda/issues/866) in my opinion.

Feedback on the API design as well as the is welcome. Integration into HA to be done, but at least it's scriptable this way.

Test:
myskoda --user="XXX" --password="YYY" set-preferred-charging ZZZ --location=loc--id=4 --enabled=False
myskoda --user="XXX" --password="YYY" set-preferred-charging ZZZ --location=loc --id=4 --enabled=True --start="12:30" --end="12:45"
myskoda --user="XXX" --password="YYY" set-preferred-charging ZZZ --location=loc --location=Zuhause --id=4 --enabled=True --start="13:40"
myskoda --user="XXX" --password="YYY" set-preferred-charging ZZZ --location=loc --location=Zuhause --id=4 --enabled=True --start="XX:ZZ" => gave expected error message

Just executed the commands above and checked the data in the MySkoda app, where the data was shown as expected.

Cheers, Max


WebSpider:
Feel free to mark your PR as ready when you are ready preparing it

myskoda/models/chargingprofiles.py
```diff
@@ -20,6 +25,11 @@ class ChargingTimes(DataClassORJSONMixin):
     start_time: time = field(metadata=field_options(alias="startTime"))
     end_time: time = field(metadata=field_options(alias="endTime"))
 
+    class Config(BaseConfig):
+        """Configuration for serialization and deserialization.."""
+
+        serialize_by_alias = True
```
dvx76, line 31: @WebSpider I think we should standardize how we handle serialization, because this is essentially a breaking change now (clients currently using `to_dict()` will get a different dict back after this change).

We're already doing this differently in different places:
1. For AC Timers we set `code_generation_options = [TO_DICT_ADD_BY_ALIAS_FLAG]`, which enables `.to_dict(by_alias=True)`
    - https://github.com/skodaconnect/myskoda/blob/630107b175633f3d40df9c34d3bc12fa72b69439/myskoda/rest_api.py#L739
    - https://github.com/skodaconnect/myskoda/blob/630107b175633f3d40df9c34d3bc12fa72b69439/myskoda/models/air_conditioning.py#L56
2. But for AC Conditions we set `serialize_by_alias = True` and just call `.to_dict()`
    - https://github.com/skodaconnect/myskoda/blob/630107b175633f3d40df9c34d3bc12fa72b69439/myskoda/models/air_conditioning.py#L118

---

IMO `.to_dict()` should return the normalized names (so not the aliases) and we should explicitly use `TO_DICT_ADD_BY_ALIAS_FLAG` + `.to_dict(by_alias=True)` whenever we need the the alias fields.

max-2468: Hi, good catch, thank you! Please let me know how you want to handle it and I‘ll adjust it accordingly.

dvx76: Please use the first option (using `TO_DICT_ADD_BY_ALIAS_FLAG`) in the new or modified code in this PR. Thank you.

PR #557, opened 2026-04-04
feat: add support for the vehicle-status widget

moledzki:
Hi,
I've added support of `/v2/widgets/vehicle-status/{vin}` endpoint, please review it.

### DIFF
--- a/myskoda/anonymize.py
+++ b/myskoda/anonymize.py
@@ -396,3 +396,17 @@ def anonymize_url(url: str) -> str:
         str: URL string with any VIN anonymized
     """
     return VIN_REGEX.sub(VIN, url)
+
+
+def anonymize_url_location(url: str) -> str:
+    """Anonymize a VIN found in a URL.
+
+    Args:
+        url: input URL string
+
+    Returns:
+        str: URL string with any VIN anonymized
+    """
+    return LOCATION_REGEX.sub(
+        f"latitude={LOCATION['latitude']}&longitude={LOCATION['longitude']}", url
+    )
--- a/myskoda/anonymize.py
+++ b/myskoda/anonymize.py
@@ -396,17 +396,3 @@ def anonymize_url(url: str) -> str:
         str: URL string with any VIN anonymized
     """
     return VIN_REGEX.sub(VIN, url)
-
-
-def anonymize_url_location(url: str) -> str:
-    """Anonymize a VIN found in a URL.
-
-    Args:
-        url: input URL string
-
-    Returns:
-        str: URL string with any VIN anonymized
-    """
-    return LOCATION_REGEX.sub(
-        f"latitude={LOCATION['latitude']}&longitude={LOCATION['longitude']}", url
-    )
--- a/myskoda/anonymize.py
+++ b/myskoda/anonymize.py
@@ -303,6 +303,7 @@ def anonymize_widget(data: dict) -> dict:
     Returns:
         dict
     """
+    data["vehicle"]["name"] = VEHICLE_NAME
     data["vehicle"]["licensePlate"] = LICENSE_PLATE
     data["vehicle"]["renderUrl"] = anonymize_url(data["vehicle"]["renderUrl"])
     data["parkingPosition"]["maps"]["lightMapUrl"] = anonymize_url_location(
--- a/myskoda/anonymize.py
+++ b/myskoda/anonymize.py
@@ -306,11 +306,14 @@ def anonymize_widget(data: dict) -> dict:
     data["vehicle"]["name"] = VEHICLE_NAME
     data["vehicle"]["licensePlate"] = LICENSE_PLATE
     data["vehicle"]["renderUrl"] = anonymize_url(data["vehicle"]["renderUrl"])
-    data["parkingPosition"]["maps"]["lightMapUrl"] = anonymize_url_location(
-        data["parkingPosition"]["maps"]["lightMapUrl"]
-    )
-    data["parkingPosition"]["gpsCoordinates"] = LOCATION
-    data["parkingPosition"]["formattedAddress"] = FORMATTED_ADDRESS
+    if "maps" in data.get("parkingPosition", {}):
+        data["parkingPosition"]["maps"]["lightMapUrl"] = anonymize_url_location(
+            data["parkingPosition"]["maps"]["lightMapUrl"]
+        )
+    if "gpsCoordinates" in data.get("parkingPosition", {}):
+        data["parkingPosition"]["gpsCoordinates"] = LOCATION
+    if "formattedAddress" in data.get("parkingPosition", {}):
+        data["parkingPosition"]["formattedAddress"] = FORMATTED_ADDRESS
     return data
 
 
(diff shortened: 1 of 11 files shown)


moledzki:
I don't know what is the best way of testing anonymize function: separate test, or another test in test_rest_api with anonymize=True?

WebSpider:
> I don't know what is the best way of testing anonymize function: separate test, or another test in test_rest_api with anonymize=True?

If i were you, i would implement it similar to the other anonymize function tests ;)

myskoda/models/widget.py
```diff
@@ -0,0 +1,51 @@
...
+
+from mashumaro import field_options
+from mashumaro.mixins.orjson import DataClassORJSONMixin
+
+from .common import BaseResponse
+from .position import ParkingCoordinates
+
+
+@dataclass
+class ChargingStatus(DataClassORJSONMixin):
```
WebSpider: I would like to prevent duplicate class names across the project, just to avoid confusion. Can you pick another name, maybe with `widget` in it?

myskoda/models/widget.py
```diff
@@ -0,0 +1,51 @@
...
+    state_of_charge_in_percent: int = field(metadata=field_options(alias="stateOfChargeInPercent"))
+    remaining_time_to_fully_charged_in_minutes: int = field(
+        metadata=field_options(alias="remainingTimeToFullyChargedInMinutes")
+    )
+
+
+@dataclass
+class Vehicle(DataClassORJSONMixin):
```
WebSpider: Same remark as line 13, please use a more unique classname

myskoda/anonymize.py
```diff
@@ -375,3 +396,17 @@ def anonymize_url(url: str) -> str:
...
+
+    Args:
+        url: input URL string
+
+    Returns:
+        str: URL string with any VIN anonymized
+    """
+    return LOCATION_REGEX.sub(
+        f"latitude={LOCATION['latitude']}&longitude={LOCATION['longitude']}", url
+    )
```
WebSpider: The point here is to anonymize the actual location in addition to the VIN, since that is PII.

myskoda/models/widget.py
```diff
@@ -0,0 +1,51 @@
...
+    name: str
+    license_plate: str = field(metadata=field_options(alias="licensePlate"))
+    render_url: str = field(metadata=field_options(alias="renderUrl"))
+
+
+@dataclass
+class VehicleStatus(DataClassORJSONMixin):
+    doors_locked: str = field(metadata=field_options(alias="doorsLocked"))
```
WebSpider: This may be an instance of `DoorLockedState`

myskoda/models/widget.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@dataclass
+class Maps(DataClassORJSONMixin):
+    light_map_url: str = field(metadata=field_options(alias="lightMapUrl"))
+
+
+@dataclass
+class ParkingPosition(ParkingCoordinates):
+    state: str
```
WebSpider: Please make this a seperate enum class, implementing `PARKED` and possibly other options.

WebSpider (changes requested):
Thanks for working on another endpoint! 👷‍♂️ 

A few inline remarks about 
- duplicate class names
- the ability to be more specific than just `str`
- location anonymization

Other than that, a great PR!

myskoda/models/widget.py
moledzki: OK

myskoda/models/widget.py
moledzki: OK

myskoda/models/widget.py
moledzki: I've changed this to `OpenState`, as it contains `CLOSED` value

myskoda/models/widget.py
moledzki: OK, I've added enum with `PARKED` and  `IN_MOTION` values

myskoda/anonymize.py
moledzki: In the Widget response we have two URLs with PII:

- `renderUrl`
- `lightMapUrl`

Vehicle render contains only VIN and map location contains only GPS coordinates. That's why I've added a separate function to anonymize just the GPS location. Should I combine these two functions into one general URL anonymizer?

moledzki:
Hi,
I've fixed most thing, the one open point is anonymize method. Should we have one method for all possible URLs or separate method for each URL type?

myskoda/anonymize.py
WebSpider: We use an anonymizer to strip PII from the response when needed, so we should have one anonymizer function per URL. If you see a way to reduce duplicate code or otherwise optimize, that is offcourse encouraged unless it brings unreasonable complexities.

moledzki: OK, I've merged these two functions into one.

myskoda/anonymize.py
```diff
@@ -389,19 +390,13 @@ def anonymize_driving_score(data: dict) -> dict:
...
 
-def anonymize_url(url: str) -> str:
-    """Anonymize a VIN found in a URL.
-
-    Args:
-        url: input URL string
+URL_REPLACEMENTS = [
+    (LOCATION_REGEX, f"latitude={LOCATION['latitude']}&longitude={LOCATION['longitude']}"),
+    (VIN_REGEX, VIN),
+]
```
WebSpider: This holds no global value, it is just a construct of 2 global variables to be used inside a function. Please move this inside `anonymize_url` if it is even needed.

myskoda/anonymize.py
```diff
@@ -410,6 +405,4 @@ def anonymize_url_location(url: str) -> str:
     Returns:
         str: URL string with any VIN anonymized
     """
-    return LOCATION_REGEX.sub(
-        f"latitude={LOCATION['latitude']}&longitude={LOCATION['longitude']}", url
-    )
+    return reduce(lambda s, args: args[0].sub(args[1], s), URL_REPLACEMENTS, url)
```
WebSpider: Why use `reduce` and not simply influence the `data` dict directly?

myskoda/anonymize.py
```diff
@@ -389,19 +390,13 @@ def anonymize_driving_score(data: dict) -> dict:
...
 
-    Returns:
-        str: URL string with any VIN anonymized
-    """
-    return VIN_REGEX.sub(VIN, url)
 
-
-def anonymize_url_location(url: str) -> str:
+def anonymize_url(url: str) -> str:
     """Anonymize a VIN found in a URL.
```
WebSpider: nit: the description doesnt filly match what the function does

WebSpider (changes requested):
Thanks a lot for keeping on pushing this. Can you move the `URL_REPLACEMENTS`, if you need them, outside the global scope please? It doesn't seem to need to be global.

myskoda/anonymize.py
moledzki: OK, I will move this variable

myskoda/anonymize.py
moledzki: OK, fixed

myskoda/anonymize.py
moledzki: OK, I can change this to iterable version.

WebSpider (approved):
Looks good to me! 👍

PR #563, opened 2026-04-17
feat: enable ansi formatting only when outputting to console

moledzki:
Hi,

Currently, it's difficult to use `grep` or `jq` to process the `myskoda` application output because it always inserts ANSI codes into the output stream.
This fix enables ANSI codes only when `myskoda` prints directly to the console.


WebSpider (approved):
Thanks for your PR!

PR #564, opened 2026-04-17
Set preferred charging times

max-2468:
Hello,

here is an updated version of the preferred charging pull request. Changes:

- switched serialization by alias to be optional for ChargingProfiles and related entites
- added test case

Best regards,
Max


WebSpider:
This replaces #550

WebSpider (commented):
Please have a look at the output from CI and apply fixes. Thanks! 👍

myskoda/cli/operations.py
```diff
@@ -508,3 +509,89 @@ async def set_aux_timer(  # noqa: PLR0913
...
+            else None
+        )
+
+        if selected_charging_times is not None:
+            if enabled is not None:
+                selected_charging_times.enabled= enabled
+            if start is not None:
+                try:
+                    #it is assumed that the car will always use it's local time,
+                    #so timezone information is not relevant
```
WebSpider: I cannot reproduce this assumption, what are you basing it on?

myskoda/cli/operations.py
```diff
@@ -508,3 +509,89 @@ async def set_aux_timer(  # noqa: PLR0913
...
+
+        selected_location = (
+            next((t for t in all_profiles.charging_profiles if t.name == location), None)
+            if all_profiles.charging_profiles
+            else None
+        )
+
+        if selected_location is None:
+            print(f"No charging location found with name {location}.")
+            return
```
WebSpider, line 562: Please also add support changing charging profiles by `ChargingProfile.id`, not only by `ChargingProfile.location`. 

Location is an optional object inside the `ChargingProfile` and defaults to `None`

WebSpider (commented):
Thanks for this addition! I noticed it is mostly based on an optional field. Could you have a look and see if you can also process a non-optional field `id`, as indicated in the remarks?

Also, it appears `ruff` is complaining about the import order. Can you do a quick `ruff check --fix`?

myskoda/cli/operations.py
max-2468: I think there's a name mismatch here. The location parameter in the cli app does not map to ChargingProfile.location (which are GPS coordinates), but to ChargingProfile.name. If the latter is empty it would likely break a lot of assumptions in the Skoda software (both in app and the car) as I cannot see a way to show the ChargingProfile to the user then. There is a validation in the iOS app for a non-empty name at least. I am happy to use a different name for the parameter, although I consider location to be more understandable here when calling the cli app.

Do you think adding support for changing the carging profiles by ChargingProfile.id (which I'd consider to be a technical primary key internal to Skoda) is still necessary?

myskoda/cli/operations.py
max-2468: The raw data in the fixtures shows the time just in HH:MM:SS format. At least the iOS app only shows HH:MM, the car as well. As the data structure is generally a pretty simple representation of what is visible in the app or the car the assumption ist that it simply shows what's there. I am not able to test whether the car would be able to do timezone calculations when to start/stop charging. I'd be suprised though. Also, the serialization of ChargingTimers formats the time in HH:MM format when sending data back to the servers.

max-2468:
The additional linting fix is done. As mentioned in the other comment, the code operates on `ChargingProfile.name` as _business primary key_. That is not an optional field in my opinion. The location field (which are GPS coordinates) is not used. I can add variantes of these methods that use `ChargingProfiles.id`, but these are not user visible anywhere when not looking at the output directly. With respect to being able to use these methods from HA at some point, being able to use the name that was given to the location by the user seems necessary, instead of the technical id.

myskoda/cli/operations.py
WebSpider: OK, we would need to document that for users of this code 👍

WebSpider:
The main reason for this, is because the HA integration is keyed by this id. 

We chose this since the user may change the name of the location, and we do not want HA to regard this as a new location (causing it to delete the old location + history and creating a new one) whenever a user changes location name.

max-2468:
Thanks, understood. I'll add/change that then. Is it sufficient to make this change to `MySkoda.set_preferred_charging_times` and leave the cli operation to be called with the location name as a parameter? That should resolve the problem for the HA integration. The cli is stateless anyway, can trivially translate from the name to id as it gets the complete current setup first anyway, and it makes more sense from a user perspective in my opinion.

WebSpider:
Yes, the cli operation is independent from the HA integration. Anything in `cli` just calls other public methods, just like HA does.

max-2468:
I've switched the library to use the `ChargingProfile.id`  instead of the name. Do you need anything else to be done here? I'd send a clean PR otherwise.

WebSpider:
Just the linter that needs some love. Probably `ruff format` will fix it.

PR #565, opened 2026-04-20
fix: added NO_UPDATE_AVAILABLE to SoftwareStatus

MarcelKilgus:
My new Enyaq returns `NO_UPDATE_AVAILABLE` for update status as apparently it's still up-to-date ;)

JSON also doesn't contain "release_notes_url" or "car_captured_timestamp", so they are returned as None by the API. Seems fine for me, so I made the unit test assert optional accordingly.

### DIFF
--- a/myskoda/models/software_status.py
+++ b/myskoda/models/software_status.py
@@ -9,6 +9,7 @@ from .common import BaseResponse, CaseInsensitiveStrEnum
 
 
 class SoftwareStatus(CaseInsensitiveStrEnum):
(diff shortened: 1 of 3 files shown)


WebSpider:
Thanks for your PR!
This is one of those cases where the content of cars can apparently wildly differ, so we really appreciate you adding your experience! :muscle:

WebSpider (approved):
lgtm

MarcelKilgus:
Cool, thanks :) I'm glad this library is around to play around with the car (and my main usage to log mileage every day). Cheers!

PR #570, opened 2026-04-27
fix: update fcm_token attribute after registering

dvx76:
Minor fix I ran into while finishing the corresponding homeassistant-myskoda updates.


WebSpider (approved):
Lgtm

PR #575, opened 2026-05-19
fix: added overall_cost field to TripStatistics

moledzki:
Hi,
 I've added missing overall_cost field to TripStatistics.


WebSpider (approved):
lgtm

PR #578, opened 2026-05-21
Trip statistic filtering

moledzki:
Hi,
I've added offset number and offset type filtering to get_trip_statistics method.

### DIFF
--- a/myskoda/cli/requests.py
+++ b/myskoda/cli/requests.py
@@ -407,9 +407,8 @@ async def trip_statistics(
     ctx: Context, vin: str, offset: int, offset_type: OffsetType, anonymize: bool
 ) -> None:
     """Print the last trip statics."""
-    print(offset)
     await handle_request(
-        ctx, ctx.obj["myskoda"].get_trip_statistics, vin, offset, offset_type, anonymize
+        ctx, ctx.obj["myskoda"].get_trip_statistics, vin, anonymize, offset, offset_type
     )
 
 
--- a/myskoda/cli/requests.py
+++ b/myskoda/cli/requests.py
@@ -7,6 +7,7 @@ import asyncclick as click
 from asyncclick.core import Context
 
 from myskoda.cli.utils import MethodArgument, handle_request, iso8601_datetime, simple_date
+from myskoda.rest_api import OffsetType
 
 if TYPE_CHECKING:
     from myskoda.myskoda import MySkoda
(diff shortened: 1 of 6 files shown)


WebSpider:
Thanks for your contribution! 

Can you please explain whats missing from the current implementation?

moledzki:
Current implementation only allows you to retrieve trip statistics for last week only. This PR allows you retrieve data from past months and weeks. 

To retrieve trip statistics for whole March pass following options to `myskoda` command: `myskoda  --user *******@gmail.com --password ********** trip-statistics --offset 2 --offset-type month TMBJ`

myskoda/cli/requests.py
```diff
@@ -365,10 +366,29 @@ async def user(ctx: Context, anonymize: bool) -> None:
...
+    default=OffsetType.WEEK.value,
+)
 @click.pass_context
-async def trip_statistics(ctx: Context, vin: str, anonymize: bool) -> None:
+async def trip_statistics(
+    ctx: Context, vin: str, offset: int, offset_type: OffsetType, anonymize: bool
+) -> None:
     """Print the last trip statics."""
-    await handle_request(ctx, ctx.obj["myskoda"].get_trip_statistics, vin, anonymize)
+    print(offset)
```
dvx76: Leftover from debugging I guess. Can you remove this `print`?

myskoda/myskoda.py
```diff
@@ -428,9 +428,27 @@ async def get_single_trip_statistics(
...
-    async def get_trip_statistics(self, vin: Vin, anonymize: bool = False) -> TripStatistics:
-        """Retrieve statistics about past trips."""
-        return (await self.rest_api.get_trip_statistics(vin, anonymize=anonymize)).result
+    async def get_trip_statistics(
+        self,
+        vin: Vin,
+        offset: int = 0,
+        offset_type: OffsetType = OffsetType.WEEK,
+        anonymize: bool = False,
+    ) -> TripStatistics:
```
dvx76, line 446: Generally don't like to add new kwargs in the middle of the signature as it can break existing callers passing unnamed arguments. E.g. someone currently using `get_trip_statistics(vin, True)` will start failing.

Ideally we force the use of keywords. But that of course would also be a breaking change 🙈 .

I know it's ugly to have `anonymize` (and `notify` for the refresh method) before the actual request parameters. But it's better than breaking someones code.

dvx76 (dismissed):
Some minor comments but generally LGTM. Thanks for your contribution!

myskoda/cli/requests.py
moledzki: OK

myskoda/myskoda.py
moledzki: I was also wondering which parameter order is the lesser evil :)

dvx76 (approved):
LGTM. Thanks for the updates!

PR #580, opened 2026-05-25
feat: Added API to allow updating the preferred charging times

max-2468:
Hi all,

the final pull request for adding the "set_preferred_charging_times" feature. This replaces #564 as it is quite a bit neater. It contains the API to change charging timers, the cli was updated to allow this as well (but using the charging profile name instead of its id).

- [x] Test case for the new API was added 
- [x] All test cases pass successfully
- [x] All automated checks are happy 
- [x] Local test runs show successful changes to the car, including when multiple charging profiles exist

Afterwards an HA action similar to what you are doing in [ha-myskoda #1074](https://github.com/skodaconnect/homeassistant-myskoda/pull/1074) would be my plan.


myskoda/cli/operations.py
```diff
@@ -508,3 +509,99 @@ async def set_aux_timer(  # noqa: PLR0913
                 print(f"No timer found with ID {timer_id}.")
         else:
             print("No AuxiliaryHeating found for the given VIN.")
+
+
+@click.command()
+@click.option("timeout", "--timeout", type=float, default=300)
+@click.argument("vin")
+@click.option("location", "--location", type=str, required=True)
+@click.option("charging_time_id", "--id", type=click.Choice(["1", "2", "3", "4"]), required=True)
```
dvx76: Can we make this a `click.IntRange` so we don't have to convert from `str` to `int` later?


```suggestion
@click.option("charging_time_id", "--id", type=click.IntRange(1, 4), required=True)
```

myskoda/cli/operations.py
```diff
@@ -508,3 +509,99 @@ async def set_aux_timer(  # noqa: PLR0913
...
+) -> None:
+    """Update preferred charging times"""
+    charging_times_id = int(charging_time_id)
+    myskoda: MySkoda = ctx.obj["myskoda"]
+
+    # As the command can be called with multiple optional parameters
+    # make sure that one item to update was actually provided before getting data from the servers
+    if enabled is None and start is None and end is None:
+        print("No information to update was provided")
+        return
```
dvx76: raise `click.UsageError` instead of printing and returning. See the main `cli()` in `myskoda/cli/__init__.py` for examples

myskoda/cli/operations.py
```diff
@@ -508,3 +509,99 @@ async def set_aux_timer(  # noqa: PLR0913
...
+
+            if end is not None:
+                try:
+                    # see above, assume the car only uses local time
+                    parsed_start = datetime.strptime(start, "%H:%M")  # noqa: DTZ007
+                    if parsed_start is not None:
+                        selected_charging_times.start_time = parsed_start.time()
+                except ValueError:
+                    print(f"Invalid end time {end}, expected HH:MM format.")
+                    return
```
dvx76: nit: we could do the time format validation at the beginning, before we start doing work like fetching charging profiles from the API, etc. Then when we get here we know `start` and `end` are either `None` or a valid `datetime`.

Even better, delegate the validation to `click` itself, e.g.

```python
class TimeType(click.ParamType):
    name = "HH:MM"

    def convert(self, value, param, ctx):
        try:
            return datetime.strptime(value, "%H:%M").time()
        except ValueError:
            self.fail(f"{value!r} is not a valid time in HH:MM format", param, ctx)

TIME = TimeType()

@click.option("start", "--start", type=TIME, required=False, default=None)
@click.option("end", "--end", type=TIME, required=False, default=None)
async def set_preferred_charging(..., start: datetime | None, end: datetime | None, ...):
```

myskoda/cli/operations.py
```diff
@@ -508,3 +509,99 @@ async def set_aux_timer(  # noqa: PLR0913
...
+@click.pass_context
+async def set_preferred_charging(  # noqa: PLR0913, C901
+    ctx: Context,
+    timeout: float,  # noqa: ASYNC109
+    vin: str,
+    location: str,
+    charging_time_id: str,
+    enabled: bool,
+    start: str,
+    end: str,
```
dvx76: These can all be `None` so would be `enable: bool | None, ...`

myskoda/myskoda.py
```diff
@@ -745,6 +745,30 @@ async def set_ac_timer(self, vin: Vin, timer: AirConditioningTimer) -> None:
         await self.rest_api.set_ac_timer(vin, timer)
         await future
 
+    async def set_preferred_charging_times(
+        self, vin: Vin, charging_profile_id: int, times: ChargingTimes
+    ) -> None:
+        """Update the proferred charging time of the vehicle at the location."""
```
dvx76: nit: typo


```suggestion
        """Update the preferred charging time of the vehicle at the location."""
```

dvx76 (commented):
Thanks for your contribution and this updated version! Generally LGTM with just a few minor comments to look into.

myskoda/cli/operations.py
max-2468: Do you want me to change it for the other occurences of this pattern (departure_timer, ac_timer, aux_heating_timer) as well in the commit with the fix for this?

dvx76: If you want to do it that would be great, thanks.

dvx76:
Thanks for the updates! LGTM but looks like you need to rebase and deal with a few merge conflicts before this can be merged 🙏

max-2468:
Hi, hopefully finally done now. I added a fix to the unit tests as they were not quite testing what they were supposed to do, even though they ran successfully.

dvx76 (approved):
Sorry for the wait. LGTM. Thanks for your work!

PR #586, opened 2026-06-03
Fix charging statistics endpoint and update response model

### LINKED ISSUE #585: charging-history not available any more
Hi there, it seems the api for the charging history was changed with an app-update 15th May - starting from that date the response is always 500-Internal Server Error - tried with rest-api / cli

aiohttp.client_exceptions.ClientResponseError: 500, message='Internal Server Error', url='https://mysmob.api.connect.skoda-auto.cz/api/v1/charging/xxxxxxxxxxxx/history?userTimezone=UTC&limit=50&from=2026-05-20T00:00:00Z&to=2026-05-20T00:00:00Z'
error: 500
message: Internal Server Error
url: https://mysmob.api.connect.skoda-auto.cz/api/v1/charging/xxxxxxxxxxxx/history?userTimezone=UTC&limit=50&from=2026-05-20T00:00:00Z&to=2026-05-20T00:00:00Z

also tried with cursor and smaller limit ?

rsa-wusel:
## Summary

Fix charging-statistics support after backend API changes.

## Changes

- add charging service specific request headers
- update charging statistics request payload to use VIN filter
- extend charging statistics response model
- support new fields:
  - applicableFilterOptions
  - missingElliConsent
  - entry id
  - title
  - primaryValue
  - secondaryValue
  - isCurveAvailable
- support updated datetime format returned by backend
- add CSV convenience helpers
- update fixture and unit tests

## Notes

The charging statistics endpoint now returns HTTP 200 again and charging sessions are correctly deserialized.

Fixes #585


WebSpider:
Can you please elaborate a bit more about how you found out about these backend changes, and where you found these API calls?

rsa-wusel:
> Can you please elaborate a bit more about how you found out about these backend changes, and where you found these API calls?

The changes were identified by comparing the current library behavior with responses returned by the latest MySkoda mobile application, combined with analysis of the current client application and its API interactions.

While investigating why charging_statistics was no longer working, I noticed that the backend response structure had changed significantly compared to the format currently expected by the library. This helped identify the updated response schema as well as additional related charging endpoints.

The implementation included in this PR was validated using actual responses from my own vehicle account and updated accordingly. The included tests are based on sanitized examples of those responses.

During the same investigation, I also identified an endpoint providing charging power curve data. However, as I have not yet completed sufficient testing and validation of that functionality, I decided not to include it in this PR. I plan to submit a separate follow-up PR once the implementation and test coverage are complete.

Since these APIs are undocumented and may change again in the future, I tried to keep the implementation as defensive as possible and based on observed payloads rather than assumptions about the backend.

myskoda/models/charging_history.py
```diff
@@ -12,10 +13,20 @@
 
 
 def _parse_local_datetime(value: str | None) -> datetime | None:
-    """Parse a user-local formatted datetime string (dd.MM.yy HH:mm) as a naive datetime."""
-    if value is None:
+    """Parse a user-local formatted datetime string."""
+    if value is None or value == "--":
```
WebSpider: Please make this a constant

myskoda/models/charging_history.py
```diff
@@ -51,7 +62,8 @@ class ChargingHistory(BaseResponse):
 @dataclass
 class ChargingStatisticsFilterOption(DataClassORJSONMixin):
     filter_type: str = field(metadata=field_options(alias="filterType"))
-    id: str
+    vin: str | None = None
```
WebSpider: We have a `Vin` type

myskoda/models/charging_history.py
```diff
@@ -120,9 +143,47 @@ class ChargingStatisticsSection(DataClassORJSONMixin):
     entries: list[ChargingStatisticsEntry] = field(default_factory=list)
 
 
+@dataclass
+class ChargingStatisticsApplicableFilterOption(DataClassORJSONMixin):
+    filter_type: str | None = field(
+        default=None, metadata=field_options(alias="filterType")
+    )
+    id: str | None = None
+    vin: str | None = None
```
WebSpider: We have a `Vin` type

myskoda/rest_api.py
```diff
@@ -139,6 +137,24 @@ def process_json(
...
         return json.dumps(anonymized)
 
+    async def _charging_headers(self) -> dict[str, str]:
+        token = await self.authorization.get_access_token()
+        return {
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/json",
+            "Accept-Language": "de-DE",
+            "X-Brand": "skoda",
+            "X-Platform": "android",
```
WebSpider: Have you tried to see if these headers are required or are you simulating the client here?

myskoda/rest_api.py
```diff
@@ -275,14 +291,23 @@ async def get_charging_statistics(
         new sessions after a certain date.
         """
         url = f"{BASE_URL_CHARGING}/charging_statistics"
-        request = ChargingStatisticsRequest(
-            started_after=start.date(),
-            started_before=end.date(),
-            selected_filter_options=[ChargingStatisticsFilterOption(filter_type="VEHICLE", id=vin)],
-        )
+        json_data = {
```
WebSpider: Can ypu see if you can compose this using existing types?

WebSpider (changes requested):
I have done a quick review without testing the code. In a next review i will test the code and see if the responses i get are compatible

myskoda/models/charging_history.py
rsa-wusel: Done. I added a module-level constant for the empty datetime placeholder and use it in the parser now.

myskoda/models/charging_history.py
rsa-wusel: Thanks, I missed that. I will switch these fields to the existing Vin type.

myskoda/models/charging_history.py
rsa-wusel: Thanks, I missed that. I will switch these fields to the existing Vin type.

myskoda/rest_api.py
rsa-wusel: I checked the headers and removed the unnecessary ones.

Why keep Accept-Language: de-DE?

The charging statistics endpoint does not return machine-readable timestamps (ISO 8601), but localized date/time strings intended for display in the mobile app.
During testing, the response format changed depending on the Accept-Language header:
de-DE -> 01.06.26, 07:11
en-*  -> 01/06/2026, 07:11
Because the returned format is language-dependent, deserialization cannot rely on a single datetime pattern. Keeping
Accept-Language: de-DE
ensures a stable and predictable response format that matches the primary parser implementation.
As an additional safeguard, support for the observed English format was added. However, the header is intentionally retained because removing it may cause the backend to return other locale-specific formats that are currently unknown and could break datetime parsing in the future.

myskoda/rest_api.py
rsa-wusel: I tried using the request dataclass serialization directly, but to_dict()/to_json() currently emits the Python field names instead of the backend’s camelCase keys. That results in a backend error.
I kept the request dataclasses as the source of the payload, but build the final JSON explicitly so the backend receives the exact expected structure.

myskoda/rest_api.py
WebSpider: Using the Accept-Language also influences the headers in the decoded CSV:

```csv
"Session ID","Begonnen am","Beendet am","Ladezeit (s)","Tatsächliche Ladezeit (s)","Gesamtenergie (kWh)","Batterieenergie (kWh)","Komfort Energie (kWh)","Start SOC (%)","Ende SOC (%)","Gesamtpreis","Energiepreis","Blockiergebühren","Verwendeter Gutscheinbetrag","Standortname","Ladestandort"
"0232fc86-2cfe-49c7-80b1-c4a1726c4770","2026-06-04T07:00:11Z","2026-06-04T07:54:24Z","3253","3244","9.00","","","22.00","34.00","","","","","",""
```

This is unwanted as the main language of this project is English by choice. Please default to English.

myskoda/rest_api.py
WebSpider: Have a look at the `myskoda/models/air_conditioning.py` and `myskoda/models/auxiliary_heating.py` files as an example.

rsa-wusel: Thanks for the pointer. I updated the request serialization to follow the existing pattern used in other models and removed the manual payload construction.

myskoda/rest_api.py
```diff
@@ -142,7 +142,7 @@ def process_json(
     async def _charging_headers(self) -> dict[str, str]:
         token = await self.authorization.get_access_token()
         return {
-            "Accept-Language": "de-DE",
+            "Accept-Language": "en-US",
             "Authorization": f"Bearer {token}",
             "X-Brand": "skoda",
             "X-Device-Timezone": "Europe/Berlin",
```
WebSpider: In other parts of the code we used, i think, GMT. Can we keep this the same overall?

rsa-wusel: Changed X-Device-Timezone from Europe/Berlin to GMT to keep it consistent with the rest of the library.

myskoda/models/charging_history.py
```diff
@@ -51,7 +72,16 @@ class ChargingHistory(BaseResponse):
 @dataclass
 class ChargingStatisticsFilterOption(DataClassORJSONMixin):
     filter_type: str = field(metadata=field_options(alias="filterType"))
-    id: str
+    vin: Vin | None = None
+    id: Vin | None = None
```
WebSpider: In the live examples I have seen, the format seems to be a bit different. Do you have both an `id` and a `vin` of format `Vin` ?

tests/test_rest_api.py
```diff
@@ -418,9 +416,18 @@ async def test_charging_statistics(
...
-    assert first.charging_start_time == datetime(2026, 5, 15, 8, 30)  # noqa: DTZ001
+    entry = all_entries[0]
+
+    assert entry.id
+    assert entry.title
+    assert entry.primary_value
+    assert entry.secondary_value
+
+    assert entry.details.session_id is not None
+    assert entry.details.charging_power_type is not None
```
WebSpider: In the class, these may both be `None`, yet you test them to not be `None`?
Same for other tests.

WebSpider (commented):
Getting there! 

Only one remark on the tests left, and one on a model that seems to have a bit different layout for me

myskoda/models/charging_history.py
rsa-wusel: I checked this again against the observed payloads.

The request payload uses `vin` inside `selectedFilterOptions`, while the response payload under `applicableFilterOptions` uses `id` and `label`.

So I kept these as separate models:
- `ChargingStatisticsFilterOption` for the request with `vin`
- `ChargingStatisticsApplicableFilterOption` for the response with `id` and `label`

tests/test_rest_api.py
rsa-wusel: Good point. I updated the test to assert the actual values from the fixture instead of only checking for non-None values.

rsa-wusel:
I also updated the datetime parsing for charging statistics.

The formattedChargingStartTime/formattedChargingEndTime values returned by the charging statistics endpoint appear to be UTC values, although the payload does not include timezone information. To avoid treating them as local naive datetimes, they are now parsed as timezone-aware UTC datetimes.

kleintonno:
Thanks for this fix — tested against a real account (Enyaq, DE) and the endpoint returns 200 again. Two findings from real-world payloads:

**1. `sessionId: "N/A"` breaks deserialization**

Public charging sessions paid via Powerpass/RFID come back with `"sessionId": "N/A"` (and `"formattedStartSoc"/"formattedEndSoc": "N/A"`). `ChargingStatisticsSessionDetails.session_id` is typed as a required `UUID`, so the whole response fails to deserialize:

```
mashumaro.exceptions.InvalidFieldValue: Field "session_id" of type UUID in
ChargingStatisticsSessionDetails has invalid value 'N/A'
```

Example `details` object (anonymized):

```json
{
  "title": "Sat 14/03/2026, 11:57",
  "locationName": "IONITY ...",
  "formattedTotalPrice": "€34.15",
  "formattedTotalEnergy": "69.7 kWh",
  "formattedStartSoc": "N/A",
  "sessionId": "N/A",
  "chargingPowerType": "DC",
  "isActiveSession": false
}
```

Suggestion: `session_id: UUID | str | None` (or keep it a plain `str`), and treat `"N/A"` like the existing `"--"` placeholder.

**2. Formatted timestamps appear to be user-local, not UTC**

`_parse_local_datetime` attaches `tzinfo=UTC`, but the backend seems to format times in the account's local timezone regardless of the `X-Device-Timezone` header. I verified this by comparing the same session in the (UTC-based) CSV export from the app with the formatted value returned by this endpoint while sending `X-Device-Timezone: GMT`: the formatted time matched Europe/Berlin local time, not GMT. With `tzinfo=UTC` all sessions are shifted by 1–2 h.

Happy to provide more payload samples if useful.

rsa-wusel:
Thanks for testing with real-world Powerpass sessions.

I've updated the model to handle sessionId: "N/A" and added a unit test covering this case.

Regarding the timestamp handling, my observations so far point in a different direction, so I'd like to gather a few more examples before changing the implementation again.

During the investigation of the charging statistics endpoint, I compared the timestamps returned by charging_statistics with the CSV export generated by the same backend. In multiple charging sessions, the values from formattedChargingStartTime and formattedChargingEndTime matched the UTC timestamps from the CSV export, even though the API response itself does not include any timezone information.

Based on those observations, I treated the values as UTC and converted them to timezone-aware datetimes to avoid silently interpreting them as naive local times.

Your findings suggest that the backend may instead return account-local time (or potentially different formats depending on charging provider, session type, or backend source). Before changing the parsing logic again, I'd like to compare a few more real-world examples.

If possible, could you provide a sanitized sample containing:

formattedChargingStartTime
formattedChargingEndTime
the corresponding CSV timestamps for the same session
your account timezone

That would help determine whether the backend consistently returns local time, UTC, or a mixture of both depending on the session.

kleintonno:
Here's the comparison you asked for — and I think it resolves the contradiction: **it's a mixture, depending on the session source.** Account timezone is Europe/Berlin (CET, UTC+1 in February).

All five February 2026 sessions, `charging_statistics` formatted values vs. the MySkoda app's CSV export (ISO 8601 UTC) for the same sessions (paired via `sessionId`):

| session type | formattedChargingStartTime | CSV start (UTC) | formattedChargingEndTime | CSV end (UTC) | offset |
|---|---|---|---|---|---|
| home AC | `06/02/2026, 17:25` | `2026-02-06T17:25:22Z` | `07/02/2026, 08:32` | `2026-02-07T08:32:23Z` | = UTC |
| home AC | `10/02/2026, 16:59` | `2026-02-10T16:59:29Z` | `11/02/2026, 06:56` | `2026-02-11T06:56:22Z` | = UTC |
| home AC | `17/02/2026, 17:39` | `2026-02-17T17:39:42Z` | `18/02/2026, 07:29` | `2026-02-18T07:29:29Z` | = UTC |
| home AC | `23/02/2026, 17:28` | `2026-02-23T17:28:49Z` | `24/02/2026, 05:29` | `2026-02-24T05:29:07Z` | = UTC |
| public (Allego, `locationName` set) | `26/02/2026, 08:49` | `2026-02-26T07:49:08Z` | `26/02/2026, 12:27` | `2026-02-26T11:27:13Z` | **+1 h = account-local (CET)** |

So your UTC observation holds for home / vehicle-recorded sessions, while public charging sessions (the ones carrying `locationName`/price data, presumably served from the charging-provider backend) are formatted in account-local time. My earlier "user-local" conclusion was based solely on the public session — sorry for the noise.

In my own client I now treat entries with `locationName` as account-local and everything else as UTC; with that, the home sessions line up with the wallbox-side timestamps to the minute. Not pretty as an API contract, but consistent within this sample. Happy to test a branch if you want to encode that heuristic (or expose the raw strings so callers can decide).

rsa-wusel:
Fix charging statistics endpoint, handle N/A session IDs, and expose UTC timestamps parsed from the CSV export for evaluation. Initial testing with real-world charging sessions indicates that the formatted JSON timestamps may represent display values with inconsistent timezone handling, while the CSV export contains explicit UTC timestamps. Further validation with additional datasets is welcome.

myskoda/models/charging_history.py
```diff
@@ -120,9 +169,84 @@ class ChargingStatisticsSection(DataClassORJSONMixin):
...
+            if started_on is None or ended_on is None:
+                continue
+
+            result[session_id] = ChargingStatisticsCsvSession(
+                session_id=session_id,
+                started_on=started_on,
+                ended_on=ended_on,
+            )
+
+        return result
```
WebSpider, line 252: All this code that is decoding and parsing the CSV seems untested. Can you add tests here?

WebSpider (changes requested):

myskoda/models/charging_history.py
rsa-wusel: Good point.

I've added tests covering the CSV helper functionality:

- Base64 CSV decoding (`csv_bytes`)
- CSV text decoding (`csv_text`)
- CSV session parsing (`csv_sessions`)
- UTC timestamp extraction from CSV rows

The tests use the existing charging statistics fixture and verify the parsed session data against the corresponding CSV entries.

This should also cover the parsing path that is used to populate the UTC timestamps from the CSV export.

myskoda/models/charging_history.py
```diff
@@ -120,9 +169,84 @@ class ChargingStatisticsSection(DataClassORJSONMixin):
...
+        if data is None:
+            return None
+
+        try:
+            return data.decode("utf-8-sig")
+        except UnicodeDecodeError:
+            return data.decode("utf-8", errors="replace")
+
+    @property
+    def csv_sessions(self) -> dict[UUID, ChargingStatisticsCsvSession]:
```
dvx76, line 224: Given all the parsing going on in here this might be a good candidate for caching (e.g. with `functools.cached_property`).

myskoda/models/charging_history.py
```diff
@@ -120,9 +169,84 @@ class ChargingStatisticsSection(DataClassORJSONMixin):
...
+            session_id = _parse_session_id(row.get("Session ID"))
+
+            if session_id is None:
+                continue
+
+            started_on = _parse_utc_datetime(row.get("Started on"))
+            ended_on = _parse_utc_datetime(row.get("Ended on"))
+
+            if started_on is None or ended_on is None:
+                continue
```
dvx76, line 244: We might want to have a debug log for skipped rows when no `session_id` or `stated_on/ended_on` are found.

dvx76 (approved):
LGTM. Comments are minor and optional (ignorable or deferrable to another PR)

@WebSpider are you still looking at this?

WebSpider (approved):
LGTM as well. We'll add the debug logging and caching as/when we need it.

PR #599, opened 2026-06-15
Add UTC helper properties and waypoint models for trip statistics

### LINKED ISSUE #594: Trip statistics endpoint mixes local times and UTC timestamps
The trip statistics API appears to return a mix of local wall-clock times and UTC timestamps within the same response.

Example from /v1/trip-statistics/{vin}/single-trips?timezone=Europe/Berlin:

Trip:

{
  "startTime": "06:15",
  "endTime": "06:57"
}

Waypoint:

{
  "departureTime": "2026-06-09T04:15:09.438Z",
  "arrivalTime": "2026-06-09T04:57:00.352Z"
}

The waypoint timestamps correspond exactly to the trip times when converted from UTC to Europe/Berlin:

04:15Z -> 06:15
04:57Z -> 06:57

Currently the library exposes these values as received from the API, but it is unclear how consumers should interpret them.

Questions:

- Should startTime/endTime be documented as local times in the requested timezone?
- Should the library expose timezone-aware datetimes where possible?
- Should helper properties be added to provide normalized UTC timestamps?

This became visible while comparing trip data with charging_statistics data, where timestamps are also returned in UTC.

rsa-wusel:
Fixes #594 by exposing the UTC timestamps already present in the trip statistics response and extending the trip model with waypoint and trip metadata.

The API currently returns a mix of local wall-clock times (startTime, endTime) and UTC waypoint timestamps (departureTime, arrivalTime). Existing fields remain unchanged for backwards compatibility, while additional UTC helper properties are populated from the waypoint data.

Changes
Added waypoint support

Introduced new models:

Coordinates
Waypoint

and added support for parsing waypoint information returned by the API, including:

coordinates
arrival/departure timestamps
state of charge information
location names and formatted addresses
distance and travel time to the next waypoint
Added additional trip metadata

The following trip fields are now exposed:

driving_time_in_min
average_electric_consumption
electric_consumption
start_battery_state_of_charge_in_percent
end_battery_state_of_charge_in_percent
start_location_name
end_location_name
is_short_trip
waypoints

The following response-level fields are also available:

short_trip_threshold_in_km
short_trip_threshold_in_mi
Added UTC helper properties

To address the mixed timezone information returned by the API, the following helper properties were added:

Trip.start_time_utc
Trip.end_time_utc

These values are populated from waypoint timestamps:

first available waypoint departureTime
last available waypoint arrivalTime

The existing fields

start_time
end_time

remain unchanged and continue to contain the local wall-clock values returned by the API.

Rest API integration

SingleTrips.populate_utc_times() is now called automatically after deserialization so the helper properties are available directly from API responses.

Example

API response:

{
  "startTime": "06:15",
  "endTime": "06:57",
  "waypoints": [
    {
      "departureTime": "2026-06-09T04:15:09.438Z",
      "arrivalTime": "2026-06-09T04:57:00.352Z"
    }
  ]
}

Results in:

trip.start_time == "06:15"
trip.end_time == "06:57"

trip.start_time_utc == datetime(2026, 6, 9, 4, 15, 9, 438000, tzinfo=UTC)
trip.end_time_utc == datetime(2026, 6, 9, 4, 57, 0, 352000, tzinfo=UTC)
Testing

Added fixture coverage for:

waypoint parsing
coordinate parsing
UTC helper property population
mixed local/UTC timestamp handling


myskoda/models/trip_statistics.py
```diff
@@ -100,10 +100,50 @@ class TripStatistics(BaseResponse):
     )
 
 
+@dataclass
+class Coordinates(DataClassORJSONMixin):
+    latitude: float | None = None
+    longitude: float | None = None
```
dvx76, line 106: Already a problem before this PR, but adding the associated models makes it a little worse: the current `anonymize_single_trip_statistics/anonymize_trip_statistics` functions are no-ops. Location details like GPS coordinates, addresses, location names, etc should be anonimized.

myskoda/models/trip_statistics.py
```diff
@@ -114,13 +154,57 @@ class Trip(DataClassORJSONMixin):
...
+    )
+    end_location_name: str | None = field(
+        default=None, metadata=field_options(alias="endLocationName")
+    )
+    is_short_trip: bool | None = field(default=None, metadata=field_options(alias="isShortTrip"))
+    waypoints: list[Waypoint] | None = field(
+        default=None, metadata=field_options(alias="waypoints")
+    )
+
+    def populate_utc_times(self) -> None:
```
dvx76, line 191: Can we enforce UTC timestamp population to happen automatically on model creation, without requiring the caller to also call this new method? E.g. using mashumaro `__post_init__`. This ensures it always happens, regardless of what creates the model instance.

myskoda/models/trip_statistics.py
```diff
@@ -1,7 +1,7 @@
-"""Models for responses of api/v2/vehicle-status/{vin}."""
```
dvx76, line 1: Models for `v2/vehicle-status` are still in here and that endpoint is still in use, right?

myskoda/models/trip_statistics.py
```diff
@@ -100,10 +100,50 @@ class TripStatistics(BaseResponse):
...
+
+
 @dataclass
 class Trip(DataClassORJSONMixin):
     id: str | None = field(default=None, metadata=field_options(alias="id"))
+    start_time: str | None = field(default=None, metadata=field_options(alias="startTime"))
     end_time: str | None = field(default=None, metadata=field_options(alias="endTime"))
+
+    start_time_utc: datetime | None = None
+    end_time_utc: datetime | None = None
```
dvx76: nit: even though these are internal/computed, it might be nice to define camelCased aliases so any json dumps by_alias have a consistent output across the fields.


```suggestion
    start_time_utc: datetime | None = field(default=None, metadata=field_options(alias="startTimeUTC"))
    end_time_utc: datetime | None = field(default=None, metadata=field_options(alias="endTimeUTC"))
```

tests/test_rest_api.py
```diff
@@ -547,6 +547,61 @@ async def test_single_trips(
         assert get_single_trip_result.vehicle_type == VehicleType.FUEL
 
 
+@pytest.fixture(name="single_trip_timezones")
```
dvx76: These new tests don't really test anything from `rest_api.py`. Can you move them into `test_trip_statistics.py`?

dvx76 (changes requested):
Sorry for the delay in getting this reviewed. Thanks for helping with this project!

Looks good overall. A couple of changes I would like to see.

rsa-wusel:
Thanks! I've addressed all review comments:

populate UTC timestamps automatically in __post_init__()
moved the model tests to test_trip_statistics.py
added aliases for the UTC helper fields
added anonymization for waypoint/location data
updated the module docstring

dvx76 (approved):
LGTM 👍
Thanks for the updates!

PR #626, opened 2026-07-14
fix: add charging_profiles to default capabilitiesthat are fetched for a vehicle

max-2468:
Although the support was added in the other commit, Charging Profiles are not actually fetched when using get_vehicles as the Capability was not listed there. To actually remove the special handling of Charging profiles in the other HA PR completely, this missing feature is still required.


dvx76 (approved):
👍 I'll release as v2.16.1 after merging

Issue #634, opened 2026-07-30
MY 27 Enyaq uses v2 API for locking and unlocking


Dielee:
Hello, 

as said here: https://github.com/skodaconnect/homeassistant-myskoda/issues/1161, the new Enyaq uses api v2 to lock and unlock the car.

PR #635, opened 2026-07-30
Update vehicle lock/unlock API version from v1 to v2

Dielee:


Dielee:
#634

WebSpider:
Thanks for your PR!

This will change the API url to v2 for all vehicles. I am hestant to change this without seeing if this works for other vehicles that support remote locking. 

Can someone with a vehicle that currently supports remote locking attempt this patch?

WebSpider:
@Dielee would you mind taking a look at the failing tests?

Dielee:
Testing fails because of "Connection refused" against v2 API for some cars. For my car it's working.

WebSpider:
We are currently looking for someone that uses this function in a vehicle other than a recent MEB vehicle (so: not fully electrical, or before MY24). If you are using this, please contact us via [Discord](https://discord.gg/t7az2hSJXq)

WebSpider:
I would be OK with this API path change if you gate it to only work for your specific model/year

dvx76:
> I would be OK with this API path change if you gate it to only work for your specific model/year

We could scope this to capability `CHARGING_MEB`? Use `v2` there and `v1` for anything else.

WebSpider:
Then we'd have to check if it works for a MY23 enyaq, for instance:

https://github.com/skodaconnect/myskoda/blob/main/fixtures%2Fenyaq_2023_5AZJJ2_80.yaml#L71

PR #637, opened 2026-08-01
fix: add missing UPDATE_AVAILABLE software update status

### LINKED ISSUE #636: Field "status" of type SoftwareStatus in SoftwareUpdateStatus has invalid value 'UPDATE_AVAILABLE'
Hi, 

My car has an update available and this is throwing an error in the log:

2026-07-31 14:39:56.817 ERROR (MainThread) [myskoda.rest_api] Failed to deserialize data: {"status":"UPDATE_AVAILABLE","currentSoftwareVersion":"5.4.4","carCapturedTimestamp":"2026-07-30T20:41:16Z","releaseNotesUrl":"https://content.vw.io/oru/skoda/eu/ota/rn-2025-10-29-rxrc/en-GB/index.html"}
  File "<string>", line 23, in __mashumaro_from_json__
  File "/usr/local/lib/python3.14/enum.py", line 709, in __call__
    return cls.__new__(cls, value)
           ~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/enum.py", line 1194, in __new__
    raise ve_exc
ValueError: 'UPDATE_AVAILABLE' is not a valid SoftwareStatus

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.14/site-packages/myskoda/rest_api.py", line 1114, in _deserialize
    data = deserialize(text)
  File "<string>", line 25, in __mashumaro_from_json__
mashumaro.exceptions.InvalidFieldValue: Field "status" of type SoftwareStatus in SoftwareUpdateStatus has invalid value 'UPDATE_AVAILABLE'
2026-07-31 14:39:56.818 WARNING (MainThread) [myskoda.myskoda] Requesting CHARGING_MEB failed: Field "status" of type SoftwareStatus in SoftwareUpdateStatus has invalid value 'UPDATE_AVAILABLE', continue

ethanhawkes-gif:
Fixes #636

A car with a pending update returns `"status": "UPDATE_AVAILABLE"` from the software-version update-status endpoint, but that value is missing from the `SoftwareStatus` enum, so deserialization fails for anyone whose car currently has an update waiting:

```
ValueError: 'UPDATE_AVAILABLE' is not a valid SoftwareStatus
mashumaro.exceptions.InvalidFieldValue: Field "status" of type SoftwareStatus in SoftwareUpdateStatus has invalid value 'UPDATE_AVAILABLE'
```

This adds `UPDATE_AVAILABLE` to the enum, placed in update-lifecycle order between `PRE_UPDATE_AVAILABLE` and `UPDATE_SUCCESSFUL`. Same shape as #619 and #607, which added other missing members of this enum as they were seen in the wild.

Added a test that deserializes the exact payload from the issue report and asserts the status parses to `SoftwareStatus.UPDATE_AVAILABLE`; it fails before the change and passes after. The full test suite passes and pre-commit (ruff, ruff-format, pyright) is clean.


WebSpider (changes requested):
Thankf for the PR! 

The test only tests for the newly added status. If we're going to cover it, we should test all statuses

ethanhawkes-gif:
Thanks for the review. I've extended the test to cover every member of `SoftwareStatus` — it's now parametrized over the enum itself rather than a hardcoded list, so any status added in future is covered automatically without editing the test.

spockster1985:
I noticed another status during update: UPDATE_IN_PROGRESS 
Since you are on it, maybe you can also add this status.

ethanhawkes-gif:
Good catch — added in e434c13. `UPDATE_IN_PROGRESS` now sits between `UPDATE_AVAILABLE` and `UPDATE_SUCCESSFUL` so the enum reads in lifecycle order, and because the test is parametrized over `SoftwareStatus` itself it picked up the new member with no test edit (6 passed, 539 in the full suite).

One thing worth a maintainer decision, which I've deliberately left out of this PR to keep it focused: this is the second missing status found in two days, and the underlying problem is that any status Škoda adds server-side raises rather than degrading. Most other enums in `models/common.py` carry an `UNKNOWN` member for exactly this reason, but `SoftwareStatus` doesn't — and `UnexpectedSoftwareUpdateStatusError` is defined in the module yet never raised anywhere. Happy to open a separate PR wiring up an `UNKNOWN` fallback (with a warning log, matching the `DoorWindowState` pattern in `models/status.py`) if you'd want that — it would stop this class of breakage from recurring. Entirely your call on whether strictness here is intentional.

WebSpider (changes requested):
Thanks for the update, a lot better now. 

Could you add the payload as a json in a seperate file, like we do for other tests as well? That would be great!

dvx76:
Woops I reviewed and merged #641 before looking at this 😅 . Sorry about that. I'll mention both in the release notes.

PR #641, opened 2026-08-02
fix(models): add missing SoftwareStatus members (UPDATE_AVAILABLE, UPDATE_IN_PROGRESS)

WAHIB-EL-KHADIRI:
## What & why

Fixes #636.

When a car has a software update available, the `software-version/update-status` endpoint returns `"status": "UPDATE_AVAILABLE"`, but `SoftwareStatus` didn't model that value. mashumaro therefore raised `ValueError: 'UPDATE_AVAILABLE' is not a valid SoftwareStatus` and the entire response failed to deserialize (reported in the issue log).

A second value, `UPDATE_IN_PROGRESS`, was reported in the same thread by another user whose car was mid-update.

## Change

- Add `UPDATE_AVAILABLE` and `UPDATE_IN_PROGRESS` to `SoftwareStatus`, placed in lifecycle order between `PRE_UPDATE_AVAILABLE` and `UPDATE_SUCCESSFUL`.
- Add `tests/fixtures/enyaq/software-version-update-available.json` using the **exact payload from the report** (status, currentSoftwareVersion, carCapturedTimestamp, releaseNotesUrl).
- Wire that fixture into the existing `test_software_updates` list so the `UPDATE_AVAILABLE` path is now covered.

## Test

`uv run pytest tests/test_rest_api.py::test_software_updates` — **1 passed** (Python 3.14). The new fixture deserializes cleanly and its `status`/`releaseNotesUrl`/`currentSoftwareVersion` are asserted by the existing test body.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #645, opened 2026-08-09
fix(models): degrade unknown enum values to UNKNOWN instead of raising

WAHIB-EL-KHADIRI:
## Problem

`CaseInsensitiveStrEnum._missing_` returns `None` for any value it doesn't recognize, so when the API introduces a status that isn't enumerated yet, `_missing_` returns `None`, the enum lookup raises `ValueError`, and the **entire response fails to deserialize**.

This has already bitten the project: #641 fixed it reactively by enumerating newly-observed `SoftwareStatus` values, and the `UnexpectedSoftwareUpdateStatusError` class in `software_status.py` (defined but never used) shows the crash was anticipated but never actually handled. Every new server-side value repeats the break.

## Fix

Generalize the existing `UNKNOWN` convention (already used across `common.py`): when a value isn't recognized, `_missing_` now falls back to the enum's `UNKNOWN` member **if it defines one**, and adds `UNKNOWN` to `SoftwareStatus`.

- Backward compatible & opt-in: enums **without** an `UNKNOWN` member keep the previous strict behavior (still raise), so nothing changes for them.
- Case-insensitive matching is unchanged and still wins over the fallback for known values.
- Other case-insensitive enums can opt into graceful degradation later by simply adding `UNKNOWN`.

## Tests

`tests/test_case_insensitive_enum.py` (4 tests, all passing):
- unrecognized value → `SoftwareStatus.UNKNOWN`
- known value still matches case-insensitively (`"update_in_progress"` → `UPDATE_IN_PROGRESS`)
- a novel `status` in JSON deserializes without raising
- regression guard: an enum without `UNKNOWN` (`ChargingState`) still raises

`ruff check` clean on the changed files.


WebSpider:
We are updating our policy on submissions to this project. Please ensure your submission is in agreement with the new Open Home Foundation AI Policy: https://developers.home-assistant.io/docs/ai_policy/

WebSpider:
@dvx76 i have always considered us throwing an error when we cant squeeze the reply into the model a strong feature, to be as deterministic as possible. 

Merging this PR would change that behaviour, i would like to know what you think about it

dvx76:
> @dvx76 i have always considered us throwing an error when we cant squeeze the reply into the model a strong feature, to be as deterministic as possible.
> 
> Merging this PR would change that behaviour, i would like to know what you think about it

Agreed. I'm actually fine with (selectively) parsing unknown values _somehow_, so that for e.g. a new SW update state doesn't break everything. But this implementation will hide the new value and would need debug output to determine what the value is. What do you think about #651 ?

WAHIB-EL-KHADIRI:
Closing in favour of #651, which is the better fix.

Preserving the raw value as a pseudo-member beats collapsing it to a single `UNKNOWN` member: the original data stays visible, the warning surfaces new API values instead of hiding them, and making it opt-in via `LenientStrEnum` keeps the strict behaviour everywhere it is still wanted. My version lost information at exactly the point where you most want it.

The tests added in #651 also cover everything I had here, plus singleton identity, logging and raw-value serialization. Nothing left to port over.

Thanks for the review time on this one.

PR #651, opened 2026-08-15
feat(models): preserve unknown enum values as pseudo-members

dvx76:
Add LenientStrEnum, a case-insensitive enum that maps unrecognized values to singleton pseudo-members carrying the raw string and logs a warning, so newly introduced API values no longer break deserialization and the original data remains visible. SoftwareStatus opts in; the dead UnexpectedSoftwareUpdateStatusError is removed. Enums without the lenient base keep the strict behavior.

Potential replacement for #645.

Fixes #641 

Note: quick PoC made with AI Agent. Need to review it myself still.

### DIFF
--- a/myskoda/models/common.py
+++ b/myskoda/models/common.py
@@ -1,11 +1,15 @@
 """Common models used in multiple responses."""
 
+import logging
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from enum import StrEnum
+from typing import cast
 
 from mashumaro import field_options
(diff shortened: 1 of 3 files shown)


WebSpider:
I think I like what this is supposed to do, but I need to look up `_missing_` and `_value2member_map` to see if there is a chance of anything getting eaten

dvx76:
Going to merge this so I can include it in the release I'm about to cut. But let me know if you have any concerns and I'll address them in another PR!

PR #652, opened 2026-08-15
test: use ids for generated test to ease identification of failures

max-2468:
This change adds readable ids to the tests generated from fixtures. It allows one to easily identify the fixture file/endpoint that is being tested instead of having to figure that out from the stack trace of the failed test.
pytest_ids


dvx76 (approved):
Nice! LGTM.

PR #640, opened 2026-08-02
fix: support one-off charging profile timers

ethanhawkes-gif:
Fixes #631.

`ONE_OFF` charging timers contain `oneOffDay` and omit `recurringOn`, but `ChargingTimers.recurring_on` was required. That made the reporter's charging-profile response fail during deserialization.

This change adds the `oneOffDay` mapping, makes the two day selectors optional, omits the unused selector on serialization, and preserves the API's `HH:MM` time format for both recurring and one-off timers.

Validation:

- 2 focused charging-profile tests passed
- 535 tests passed
- Ruff and Ruff format passed on the changed files
- Pyright passed
- `uv lock --check` passed

One theoretical edge remains: a malformed timer with neither day selector can deserialize, but its by-alias time serialization falls back to `HH:MM:SS`. I did not find this shape in the issue payload or open work; happy to enforce that invariant here if preferred.

PR #653, opened 2026-08-15
fix/feat:Add support for One Off Charging Timeres as well as start_climatisation fix

max-2468:
While doing some tests regarding PR #640 I did stumble across the toggle for start the climatisation of the car as well. Switching that on made a new field appear in the data send by the server.

Apparently the app/server/car will only notice this field once it has been manually set once, and will default a missing value to False. To handle this new field while ensuring compatiblity to the behaviour shown by the app (and not having to manually edit a lot of fixtures as well as ensuring compatibility for new fixtures) a special type was used to handle the distinction True, False, Unknown/None.

The handling of One_off charging timers is based on PR #640 by ethanhawke-gif.

This PR includes:

- new fixtures testing all options available on charging profiles/charging timers, one_off timers as well as the climatisation switch
- an updated handling of fixture test cases to handle the optionality of new fields (by doing a complete JSON => model => JSON roundtrip and then loading both the resulting JSON as well as the fixture as a dict).
- some more fixes to the charging profile test case


myskoda/models/chargingprofiles.py
```diff
@@ -10,12 +10,26 @@
     BaseConfig,
 )
 from mashumaro.mixins.orjson import DataClassORJSONMixin
+from mashumaro.types import SerializationStrategy
+from tribool import Tribool
```
dvx76, line 14: Can you try to implement this without the tribool package.

This adds another package dependency for something relatively trivial. More importantly that package hasn't been updated since 2016 and includes some questionable behavior (e.g. `bool(Tribool(True))` raises a `TypeError`)

The `start_climatisation` field can just be typed as `bool | None`.

dvx76 (changes requested):
Thanks for your contribution!

Can you please look at replacing the `tribool` package?

For the `one_off_day` fix, either remove that from your PR, or rebase on top of #640 . Don't create your own commit with the same change.

PR #654, opened 2026-08-15
fix: refresh vehicle data on vehicle MQTT events

### LINKED ISSUE #1170: Not getting timely updates when car is parked/locked
### Please confirm:

- [x] I have searched the existing open and closed issues this problem
- [x] I have updated the integration to the latest available version

### Home Assistant installation and version

2026.07

### Car model

Elroq 85

### Integration version

1.35.0

### Latest working release

_No response_

### Bug Description

When I stop the car, lock and walk away, the official app seems to update in a timely manner with the location, mileage etc, however home assistant integration seems to only get that info on the next planned polling interval (default 30 mins). I understand MQTT updates should happen, but dont seem to be receiving them? I've enabled debug logs for MQTT and myskoda and below is the startup messages with a few polled attempted (set to every 2 mins for now until this issue is figured out).

I will provide logs later today when I return from a drive to show whats going on (will reset to 30 min polling too).

dvx76:
Vehicle events such as vehicle-connection-offline or vehicle-ignition-status-changed were parsed but never acted upon, leaving clients with stale data until the next scheduled poll.

Trigger a refresh_vehicle() for all VehicleEvent types. The refresh is debounced so bursts of events are coalesced into at most one immediate and one trailing refresh.

Fixes skodaconnect/homeassistant-myskoda#1170

### DIFF
--- a/myskoda/models/event/__init__.py
+++ b/myskoda/models/event/__init__.py
@@ -31,6 +31,7 @@ from .service import (
     ServiceEventOdometer,
 )
 from .vehicle import (
+    VehicleEvent,
     VehicleEventAwake,
     VehicleEventConnectionOffline,
     VehicleEventConnectionOnline,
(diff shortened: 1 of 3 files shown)


WebSpider (approved):
LGTM!

PR #660, opened 2026-09-04
docs: clarify EOL migration guidance for new official API

Nortaq-PlayNexus:
Hello skodaconnect team - thank you for maintaining myskoda and for the transparent EOL announcement.

This small docs-only change adds a concise Migration tip callout directly below the existing EOL banner:
- points new users to the official public API docs (https://public.api.connect.skoda-auto.cz/docs)
- reassures existing Home Assistant users that security/compat fixes continue until Oct 2026
- suggests pinning myskoda<3 for a stable surface during migration

Motivation: new visitors landing on PyPI/README see the EOL warning but may miss the next step. This keeps the original banner intact and just adds one helpful paragraph in the existing tone.

No code, no dependencies, no behavior change - purely documentation. Happy to adjust wording or placement if you prefer.

Thanks for considering!


dvx76 (changes requested):
Hi,

Thanks for your PR. I like the initiative but I'm a little confused about the wording and discrepancies between the PR description and the actual additions.

> Reassures existing Home Assistant...

This library is independent of HA and most HA users never see it. HA communication happens in skodaconnect/homeassistant-myskoda

> Pin `myskoda<3` if you need a stable API surface during migration.

I don't understand what this tries to convey. There is no migration and there will never be a `myskoda==3`. A new library package is release to interact with the new API: https://pypi.org/project/skoda-public-api/

